PR #2, opened 2026-01-16
Fix isSecurejoinMessage, in order to prevent cleartext mail from getting through

Hocuri:
This PR is partly vibe-coded, but I read all the code, and fixed things that were wrong.

Also, I tested that everything works now.


internal/pgp_verify/pgp_verify.go
```diff
@@ -68,11 +68,12 @@ func IsSecureJoinMessage(header textproto.Header, body io.Reader) bool {
 		return false
 	}
 
-	if mediatype != "multipart/mixed" && mediatype != "multipart/alternative" {
+	// Only accept multipart/mixed
```
adbenitez: this comment is kind of pointless/too-obvious

adbenitez (approved):

adbenitez (approved):

PR #5, opened 2026-01-18
Feature: Add JIT Registration Flag to Control Automatic Account Creation

aleskxyz:
## Problem

The `__REGISTRATION_OPEN__` flag controlled both:
- The `/new` API endpoint and web interface account creation
- Automatic account creation during IMAP/SMTP login attempts  
- Automatic account creation during email delivery to non-existent recipients

This coupling made it impossible to allow web-based account creation while preventing automatic account creation on login/delivery. For chatmail servers, this is a common requirement: users should be able to create accounts via the web interface, but accounts should not be automatically created when someone tries to log in with random credentials or when spam is delivered.

## Solution

Introduce a separate `__JIT_REGISTRATION_ENABLED__` (Just-In-Time registration) flag that controls automatic account creation independently from the general registration flag. This allows fine-grained control over when accounts are created.

### Core Implementation

**Framework (`framework/module/auth.go`)**
- Added `IsJitRegistrationEnabled()` and `SetJitRegistrationEnabled()` methods to `PlainUserDB` interface

**Authentication (`internal/auth/pass_table/table.go`)**
- Changed `AuthPlain()` to check JIT flag instead of registration flag when user doesn't exist
- Added `IsJitRegistrationEnabled()`: checks `__JIT_REGISTRATION_ENABLED__` flag, defaults to `__REGISTRATION_OPEN__` if not set
- Added `SetJitRegistrationEnabled()`: sets the JIT flag in settings table (or main table if no settings table)

**Storage (`internal/storage/imapsql/`)**
- Updated delivery hook (`delivery.go`) to check JIT flag instead of registration flag for auto-creation
- Updated `GetOrCreateIMAPAcct()` (`imapsql.go`) to respect JIT flag - only auto-creates if JIT enabled, otherwise returns error for non-existent users
- Added `IsJitRegistrationEnabled()` method to storage module

### CLI Commands (`internal/cli/ctl/users.go`)

Added new `maddy creds jit` subcommands:
- `maddy creds jit enable` - Enable automatic account creation on login/delivery
- `maddy creds jit disable` - Disable automatic account creation (only `/new` API can create accounts)
- `maddy creds jit status` - Show current JIT registration status

### Web Interface (`internal/endpoint/chatmail/`)

**Backend (`chatmail.go`)**
- Added `JitRegistrationEnabled` field to template data structures
- Passes JIT status to web interface templates

**Frontend (`www/index.html`)**
- **Major refactor of account generation logic:**
  - **JIT enabled**: Client-side random generation (instant, no server round-trip)
    - Generates username/password in browser
    - Creates `dclogin:` link immediately
    - No loading time, works offline
  - **JIT disabled**: Server-side `/new` API call
    - Shows loading indicator during account creation
    - Handles errors gracefully with user-friendly messages
    - Account is created on server before QR code is shown
- Improved Delta Chat protocol handler detection:
  - Detects if Delta Chat app is installed
  - Shows helpful error message if protocol handler doesn't exist
  - Uses blur/visibility events and timeout fallback
- Better UX:
  - Loading indicators
  - Error message display
  - Button state management
  - Auto-generation only when both registration open AND JIT enabled

### Default Behavior
- If `__JIT_REGISTRATION_ENABLED__` is not explicitly set, it defaults to `__REGISTRATION_OPEN__`
- This ensures backward compatibility - existing configurations continue to work

### Flag Storage
- Stored in settings table (if configured) or main credentials table
- Follows same precedence rules as other flags (`__REGISTRATION_OPEN__`, `__TURN_ENABLED__`, etc.)

## Use Cases

**Scenario 1: Web interface only (recommended for chatmail)**
```bash
maddy creds registration open
maddy creds jit disable
```
- Users can create accounts via web interface (`/new` API works)
- Accounts are NOT auto-created on login attempts
- Accounts are NOT auto-created on email delivery
- Prevents spam account creation and unauthorized access

**Scenario 2: Full automatic registration**
```bash
maddy creds registration open
# JIT defaults to registration status, so auto-creation works
```
- Both web interface and automatic creation work
- Accounts created on login/delivery

**Scenario 3: Registration closed**
```bash
maddy creds registration close
# JIT defaults to registration status (false)
```
- No account creation via any method

### When Accounts Are Auto-Created

Accounts are automatically created when:
1. **IMAP/SMTP login** (`AuthPlain()`): User doesn't exist AND JIT is enabled
2. **Email delivery** (`delivery.go`): Recipient doesn't exist AND JIT is enabled  
3. **IMAP access** (`GetOrCreateIMAPAcct()`): User doesn't exist AND JIT is enabled

### Web Interface Account Generation

- **JIT enabled**: Client generates random credentials → instant QR code → account created on first login
- **JIT disabled**: Client calls `/new` API → server creates account → returns credentials → QR code shown

The web interface automatically chooses the appropriate method based on JIT status.

## Backward Compatibility

✅ **Fully backward compatible**
- Existing configurations work without changes
- JIT flag defaults to registration status if not set
- No breaking changes to existing behavior

## Files Changed

- `framework/module/auth.go` - Interface definition
- `internal/auth/pass_table/table.go` - JIT flag implementation
- `internal/cli/ctl/users.go` - CLI commands
- `internal/endpoint/chatmail/chatmail.go` - Template data
- `internal/endpoint/chatmail/www/index.html` - Web interface logic
- `internal/storage/imapsql/delivery.go` - Delivery hook
- `internal/storage/imapsql/imapsql.go` - Storage JIT check

### DIFF
--- a/docs/chatmail-setup.md
+++ b/docs/chatmail-setup.md
@@ -455,6 +455,21 @@ Let's Encrypt certificates renew automatically, but you may need to reload maddy
 sudo systemctl reload maddy
 ```
 
+### Managing JIT Registration
+
+JIT registration controls automatic account creation during login attempts and email delivery:
+
+```bash
+# Enable automatic account creation
+sudo maddy --config /etc/maddy/maddy.conf creds jit enable
+
+# Disable automatic account creation
+sudo maddy --config /etc/maddy/maddy.conf creds jit disable
+
+# Check JIT registration status
+sudo maddy --config /etc/maddy/maddy.conf creds jit status
+```
+
 ## Troubleshooting
 
 ### Common Issues
--- a/docs/chatmail/authentication.md
+++ b/docs/chatmail/authentication.md
@@ -12,12 +12,12 @@ The core logic resides in the `auth.pass_table` module (`internal/auth/pass_tabl
 3.  **Branching Logic**:
     *   **User Exists**: Standard password verification is performed (Bcrypt hash comparison).
     *   **User Does Not Exist**:
-        *   The system queries the registration status (dynamically stored in the `settings` table).
-        *   If `__REGISTRATION_OPEN__` is `true`:
+        *   The system queries the JIT registration status (`__JIT_REGISTRATION_ENABLED__`).
+        *   If `__JIT_REGISTRATION_ENABLED__` is `true`:
             *   A new user entry is created immediately.
             *   The provided password is hashed and stored.
             *   The login attempt is granted.
-        *   If `__REGISTRATION_OPEN__` is `false`:
+        *   If `__JIT_REGISTRATION_ENABLED__` is `false`:
             *   The login attempt is rejected with `Invalid Credentials`.
 
 ## Technical Implementation Details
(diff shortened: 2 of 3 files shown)


omidz4t:
The Delta Chat mechanism works in a way where relays are only used as receivers and temporary storage for messages until the recipient comes and picks them up.

The username and password for SMTP and IMAP are actually simple credentials that grant access to that mailbox (as a form of trust). However, by default, all messages must be encrypted. The account creation mechanism works like this: if you send a message to the server with a new username for the first time and the account doesn't exist, the server accepts your message as long as registration is open. You can find more details in [docs/chatmail/authentication.md](https://github.com/themadorg/madmail/blob/main/docs/chatmail/authentication.md).

I agree that just closing registration is not enough; admins should be able to manually create accounts for family members. Until we have a WebUI for management, I think it's better to do this via the command line. When the server registration is closed, we can provide a temporary account using a specific command.

aleskxyz:
> Hmm, interesting idea, but we should think about it more. The whole concept of Delta Chat relies on the fact that anyone can log in with any username/password, so if they lose their account, they can easily get back in and start messaging again.
> 
> But what you're saying also makes sense. Thanks for the input; either way, an extra feature doesn't hurt anyone, and I think we can have this as an option.

Yes, this is the trade-off, the admin has to keep the username and password persistent and secure. I hope to see this change merged soon, as it is not a breaking feature and is backward compatible.

Thanks

PR #26, opened 2026-01-23
fix(db): implement in-memory SQLite support with background syncing

omidz4t:
## Description
Implemented in-memory SQLite support with periodic background syncing to disk. This feature significantly improves performance for high-traffic servers by running the database in RAM while ensuring data persistence through periodic snapshots using SQLite's `VACUUM INTO` command.

#### 1. Core Database Engine (`internal/db/db.go`)
*   **In-Memory DSN**: Added logic to use `file::memory:?cache=shared` when `InMemory` is enabled.
*   **Initial Data Load**: Implemented `loadFromDisk` which uses SQLite's `ATTACH` mechanism within a single connection to safely copy data from the disk file to memory at startup.
*   **Background Synchronizer**: Created a background loop that uses `VACUUM INTO` to create atomic, consistent snapshots of the memory database to the disk path every interval.
*   **SyncLock GORM Plugin**: Developed a custom GORM plugin that hooks into `Before` and `After` callbacks (Create, Update, Delete, Query, etc.) to manage a `RWMutex`. This prevents data corruption by pausing DB activity during the brief window of disk synchronization.
*   **CLI Bypass**: Added a check for `module.NoRun` to ensure CLI tools (which don't run the full server) interact directly with the disk file instead of launching a separate memory instance.

#### 2. Storage & Table Integration
*   Updated `imapsql` storage and and all table modules (`gorm`, `sql_query`, `sql_table`) to support and forward `sqlite_in_memory` and `sqlite_sync_interval` configuration parameters.

#### 3. Installer & Templates (`internal/cli/ctl/`)
*   **CLI Flags**: Added `--sqlite-in-memory` and `--sqlite-sync-interval` to the `install` command.
*   **Interactive Setup**: Added a "SQLite Performance Configuration" section to the interactive installer.
*   **Config Template**: Updated `maddy.conf.j2` to conditionally include the new performance directives.

#### 4. Chatmail Endpoint Integration
*   Added support for in-memory mode for the sharing and registration databases.
*   Updated the internal documentation (`database_docs.html`) with instructions for system administrators.

#### 5. Documentation
*   Created `docs/chatmail/database.md` with a detailed architectural overview.

## Type of Change
- [x] Bug fix (non-breaking change which fixes an issue)
- [x] Documentation update

## AI Responsibility

## Testing
- [x] I have run `make test` and all tests passed.

### DIFF
--- a/docs/chatmail/database.md
+++ b/docs/chatmail/database.md
@@ -1,0 +1,59 @@
+# Madmail Database Architecture
+
+This document describes the database architecture of Madmail, including the in-memory SQLite support and synchronization mechanisms.
+
+## Overview
+
+Madmail uses [GORM](https://gorm.io/) as its ORM, providing a driver-agnostic layer for database operations. It supports SQLite, PostgreSQL, and MySQL.
+
+### Core Components
+
+- **Quota Management**: Handled via the `quotas` table, tracking user storage limits and account creation dates.
+- **Contact Sharing**: Handled via the `contacts` table for shareable DeltaChat invitation links.
+- **Generic Key-Value**: Modules like `sql_table` use the `TableEntry` model for simple storage.
+
+## In-Memory SQLite Support
+
+To enhance performance, Madmail supports running SQLite databases entirely in RAM. This significantly reduces disk I/O latency for high-traffic servers.
+
+### How it Works
+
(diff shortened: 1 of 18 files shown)


madolg:
outdated. we can revisit if needed.

PR #32, opened 2026-02-12
fix: in non-interactive installs, make --ip work and autogenerate TURN secret

j4n:
## Description

- Fix: **the TURN secret was empty in non-interactive mode** because secret generation was inside runInteractiveConfig(), skipped in non-interactive mode, and ctx.IsSet("enable-turn") returns false for default values. Extract secret generation into ensureRequiredSecrets() called for both modes. 

- Fix: **--ip was only working in --simple mode**, change if-structure so that in advanced mode: --ip now sets PublicIP and A record. This allows automated non-interactive installations to have working TURN/Iroh that require PublicIP configuration.

## AI Responsibility

## Testing
- Deltachat integration tests applicable because this affects the installer
- Tested various combinations manually and through deploy automation.

### DIFF
--- a/internal/cli/ctl/install.go
+++ b/internal/cli/ctl/install.go
@@ -1535,9 +1535,9 @@ WantedBy=multi-user.target
 		}
 
 		if !config.SkipSystemd {
-			exec.Command("systemctl", "daemon-reload").Run()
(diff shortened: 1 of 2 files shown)


omidz4t:
Thanks a lot for your contribution :)

PR #55, opened 2026-04-17
feat: deliver messages via hardlinks to recipients on the same server

madolg:
implements efficient message delivery for multiple recipients on the same server using filesystem hardlinks.


omidz4t:
Thanks @madolg for this — hardlinking duplicate deliveries on the same host is a clean way to cut disk I/O and space without changing protocol semantics. Appreciate the clear focus on efficiency for multi-recipient local delivery.

PR #62, opened 2026-05-14
fix: get rid of msgscount in go-imap-sql.

NotExactlySiev:
## Description
The go-imap-sql module is incomplete and riddled with concurrency bugs and performance issues, which isn't surprising considering it's in beta and doesn't seem to get much attention or maintenance. I administrate a Deltachat relay of ~40 people, and the postgresql database is already having issues locking up when the traffic is high, often completely hanging all message delivery every 10 minutes or so for several minutes at a time.

I've identified some of the causes of this, and also some other serious issues with the SQL code (external message bodies not being reference counted properly, staying around even after they're not referenced anymore and clogging up the database and disk space) but for this patch I'm focusing on the completely unused `msgsCount` column. Not only is this column never read, and its high frequency updating has been one of the main sources of deadlocks, but due to concurrency bugs the value stored in it isn't even correct in the first place.

## Type of Change
- [x] Bug fix (non-breaking change which fixes an issue)
- [ ] New feature (non-breaking change which adds functionality)
- [ ] Breaking change (fix or feature that would cause existing functionality to not work as expected)
- [ ] Documentation update

## AI Responsibility

## Testing
- [ ] I have added a new test scenario in `tests/deltachat-test/scenarios/`.
- [ ] I have run `make test` and all tests passed.
- [x] I have run `make test-unit` and all unit tests passed.

### DIFF
--- a/internal/go-imap-sql/backend.go
+++ b/internal/go-imap-sql/backend.go
@@ -156,6 +156,7 @@ type Backend struct {
 	setSubbed          *sql.Stmt
 	uidNextLocked      *sql.Stmt
 	uidNext            *sql.Stmt
+	uidNextIncrease    *sql.Stmt
 	hasChildren        *sql.Stmt
 	uidValidity        *sql.Stmt
 	msgsCount          *sql.Stmt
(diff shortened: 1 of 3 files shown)


omidz4t:
Thanks for the PR. I'm reviewing it.

I want database usage to be as minimal as possible, and your change is a step in that direction. If you have time, I'd be grateful for more help spotting similar hotspots elsewhere.

Longer term, the idea is to keep things like message counts and federation-server state in memory, sync to the DB when the server is quiet, and sync from the DB on startup. What you did aligns well with that.

Thanks again. Looking forward to your further contributions.

omidz4t:
the problem is fixed in the new version of madmail (madmailv2) 
I'm greatly appreciate your pull request but for now I do close it because v2 is complete rewrite and I'm planing to put effort on that.
:) good things will come soon.

PR #76, opened 2026-07-07
perf(delivery): parallelize federated group fan-out (#24)

### LINKED ISSUE #24: [BUG] High latency in federated group messages
# 📋 Overview
Investigate and resolve the significant latency observed when delivering group messages to users on remote servers (federated servers), while local delivery remains fast.

## 👤 User Story / Motivation
- **As a:** Madmail User
- **I want to:** Receive messages in federated groups as quickly as I do in local groups.
- **So that:** The group chat experience remains seamless regardless of where the members' accounts are hosted.

---

## 🔍 Observed Behavior
- Messages sent to users on the same server are delivered almost instantly.
- Messages sent to users on remote federated servers experience noticeable delays.

---

## 🛠 Investigation Strategy
- **Queue Analysis:** Check maddy's outgoing queue management and retry logic.
- **Network Tracing:** Measure latency between Madmail and typical federated mail servers.
- **Concurrency:** Verify if outgoing SMTP deliveries are properly parallelized or if they are being bottlenecked.
- **DNS/MX Lookup:** Investigate if MX record resolution or connection establishment to remote servers is causing the overhead.

---

## ✅ Acceptance Criteria (AC)
- [ ] Root cause of the federation delay identified.
- [ ] Optimization implemented to speed up outgoing deliveries.
- [ ] Latency for federated messages reduced to acceptable levels (comparable to standard SMTP delivery times).

---

## 🧪 Verification Plan
- [ ] **Scenario Test:** Use `test_07_federation.py` or a new dedicated scenario to measure delivery time across two local containers simulating different domains.
- [ ] **Log Analysis:** Audit logs for any timing information during the SMTP `RCPT TO` and `DATA` phases for remote targets.

---

## 📅 Roadmap/Checklist
- [ ] Conventional Commits used.
- [ ] Definition of Done reached.

mamaru-a:
## Summary

Fixes #24 — high latency delivering group messages to recipients on remote (federated) servers, while local delivery stays fast.

## Root cause

The two group fan-out paths did not follow the same principle:

- **Local group delivery** (`deliver_local_messages`) follows the madmail principle: write the message body **once**, then `hard_link` it for every other recipient — same bytes, one inode, no second copy.
- **Federated group delivery** walked recipients one at a time, writing a **full separate copy** of the body (`.body`, fsync'd) per recipient, inline in the SMTP submission transaction and in `route_message`. So an N-recipient federated group cost N full body writes + 2N fsyncs, serialized, while a local group cost one write + N-1 cheap hard links. That asymmetry is the "local fast, federated slow" report.

## Fix

Apply the same write-once + hard-link principle to the outbound queue:

- `QueueStore::write_shared` — writes the body to disk **once**, hard-links that single inode into every other recipient's `.body` entry, and writes a small per-recipient `.meta`.
- `OutboundQueue::enqueue_batch` uses it and dispatches each entry to the worker once durable.
- `DeliveryContext::enqueue_remote_batch` wires it into the SMTP submission path (`session.rs`) and the inbound router (`route_message`), which now collect remote recipients and enqueue them as one batch.

A federated group of N recipients now costs **one body write + N-1 hard links** instead of N full copies.

**Durability / removal unchanged:** entries share one inode, so `remove` on a delivered recipient drops a single hard link; the bytes survive until the last recipient's entry is removed — refcount-by-hard-link, exactly like local delivery. All entries are fsync'd before `250 OK`.

## Testing

- `enqueue_remote_batch_writes_every_recipient`: fans out to 25 remote recipients, then asserts (a) every recipient has a durable queue entry and (b) all `.body` files resolve to a **single inode** with link count == recipient count — proving the body is written once.
- `cargo test -p chatmail-delivery` (15 passed), `cargo test -p chatmail-smtp` (29 passed), `cargo clippy` clean.

## Verification plan (per issue)

`tests/deltachat-test/scenarios/test_07_federation.py` exercises cross-server vs. same-server delivery and can measure the delivery-time gap closing for multi-recipient federated groups.


crates/chatmail-delivery/src/queue/store.rs
```diff
@@ -92,6 +92,57 @@ impl QueueStore {
...
+        next_attempt_unix: u64,
+    ) -> Result<Vec<String>> {
+        let now = now_unix();
+        let ids: Vec<String> = rcpts
+            .iter()
+            .map(|_| uuid::Uuid::new_v4().to_string())
+            .collect();
+
+        // 1. Write the body exactly once (the only real copy on disk).
+        self.write_body(&ids[0], body).await?;
```
KamyarAT, line 151: ### No rollback in `write_shared`
   Mid-batch failure can leave:
   • orphan `.body` files (no `.meta` → invisible to `list_ids` / `purge_all`), or
   • partial complete entries on disk while SMTP returns error → on restart `reload_disk_queue` delivers them; client retry can **duplicate** those recipients.

crates/chatmail-delivery/src/queue/store.rs
```diff
@@ -92,6 +92,57 @@ impl QueueStore {
...
+    /// recipient order.
+    ///
+    /// Because the entries share one inode, [`Self::remove`] unlinking a delivered
+    /// entry just drops one link; the body's bytes live until the last recipient's
+    /// entry is removed — the same refcount-by-hard-link that local delivery relies
+    /// on.
+    pub async fn write_shared(
+        &self,
+        mail_from: &str,
+        rcpts: &[String],
```
KamyarAT, line 115: ### Empty `rcpts` panics 
   Public API indexes `ids[0]` with no empty guard. `enqueue_batch` guards empty, but `write_shared` does not.

KamyarAT (changes requested):

PR #77, opened 2026-07-08
fix: deliver /mxdeliv federation mail to all X-Mail-To recipients

experintellia:
## Problem

`/mxdeliv` delivers an inbound federated message to only the **first** `X-Mail-To` recipient. The federation protocol ([TDD 07-federation.md](docs/TDD/07-federation.md)) defines one POST per message with one `X-Mail-To` header **per recipient**, and requires "Deliver to **each** recipient's mailbox" (200 = "Accepted for **all** valid recipients"). But `handle_mxdeliv` reads the recipient via `HeaderMap::get()`, which returns only the first header value:

```rust
let rcpt = header_str(headers, "x-mail-to")
    .ok_or_else(|| ChatmailError::protocol("missing X-Mail-To"))?;
```

Everything downstream — domain/account validation, quota, `write_blob`, notify — is single-recipient. Every further `X-Mail-To` is silently ignored, and since the sender still gets a `200`, the message is never retried for the dropped recipients: it is silently lost. Nothing else in the codebase parses multiple `X-Mail-To` values (`get_all` appears nowhere); the test helper in `tests/deltachat-test/scenarios/test_22_mxdeliv_security.py` even builds multi-header requests ("Add each X-Mail-To as a separate header"), but no test asserted that all of them are delivered. The SMTP session path (`chatmail-smtp/src/session.rs`) fans out correctly — only the federation HTTP path is affected.

## Visible symptom (how this was found)

A Delta Chat group containing **two accounts on the same madmail server** plus members on other servers: every group message from a federated member arrives with two `X-Mail-To` headers, so only the first-listed local account receives it — deterministically. For the other account the group appears to only ever get messages from local members. 1:1 chats (single recipient) and groups with just one local account work fine, which makes this very confusing to debug from the client side.

Reproduce against a server: POST to `/mxdeliv` with two `X-Mail-To` headers for two local accounts and a PGP-encrypted body — only the first account's maildir gets the message; swap the header order and the other one gets it.

## Fix

- Parse **all** `X-Mail-To` headers (`get_all`); `400` only when there are none.
- Validate each recipient individually, keeping the existing silent-drop semantics per recipient (non-local domain, missing account / reserved localpart).
- Check quota per recipient; an over-quota recipient is skipped with a warning and only fails the request (`507`) when **no** deliverable recipient remains — erroring for all would make the remote queue retry and re-deliver the message to recipients that are fine.
- Fan out through the existing `chatmail_storage::deliver_local_messages` (hardlink fan-out, partial-failure outcome), mirroring the SMTP session path: quota is charged and notifications sent only for recipients whose body is durably on disk; per-recipient failures are logged.

Single-recipient requests behave exactly as before (same check order, same status codes).

## Test

`p7_delivers_to_all_x_mail_to_recipients`: one POST with three `X-Mail-To` headers (two existing users + one unknown) must deliver to both users and skip the unknown one without affecting them. Fails on `main` (second recipient gets nothing), passes with this change. `cargo test -p chatmail-fed`: 12 passed.

### DIFF
--- a/crates/chatmail-fed/src/mxdeliv.rs
+++ b/crates/chatmail-fed/src/mxdeliv.rs
@@ -24,7 +24,7 @@ use axum::response::IntoResponse;
 use chatmail_db::{is_federation_sender_blocked, DbPool};
 use chatmail_pgp::{enforce_encryption, EnforceOptions};
 use chatmail_state::AppState;
-use chatmail_storage::write_blob;
+use chatmail_storage::deliver_local_messages;
 use chatmail_types::ChatmailError;
 
 use crate::security::recipient_matches_server;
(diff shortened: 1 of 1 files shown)


omidz4t:
Thanks a lot :heart_eyes:

PR #136, opened 2026-08-03
feat(versions): version manager, update latest, and versions CLI

omidz4t:
## Summary

- Add platform version tree (`/opt/madmail` on Unix, `%ProgramFiles%\Madmail` on Windows) with `MADMAIL_INSTALL_ROOT` override.
- New CLI: `madmail versions list [--remote] | current | use | prune | remove | path` (JSON supported).
- `versions use` always re-verifies Ed25519 before switching.
- `madmail update latest` / `upgrade latest` resolves GitHub Releases latest asset and uses the existing download pipeline (size caps, archive rules, mandatory signature, preflight).
- Successful upgrades archive into the version tree and set the active pointer when the install root is writable.
- TDD: `docs/TDD/24-version-manager.md`.

## Security

- No private signing keys are included; only the existing public verify key remains in-tree.
- GitHub “latest” is not a trust shortcut — signature verification still runs.

## Test plan

- [x] `cargo test -p chatmail version_manager` (35)
- [x] `cargo test -p chatmail upgrade::`
- [x] `cargo test -p chatmail-config --lib cli::`
- [x] Docker smoke on Ubuntu 24.04: list/use(unsigned reject)/prune/remove, `list --remote`, `update latest` (sig + archive under install root)
- [ ] CI green
- [ ] Manual root install path on a real host (optional)

### DIFF
--- a/crates/chatmail/src/ctl/versions_cmd.rs
+++ b/crates/chatmail/src/ctl/versions_cmd.rs
@@ -146,11 +146,30 @@ fn use_cmd(args: &Args, root: &std::path::Path, version: &str) -> Result<()> {
     let out = CtlOut::from_args(args, "versions use");
     let id = version_manager::sanitize_version_id(version)?;
     let bin = version_binary_path(root, &id);
-    if !bin.is_file() {
-        return Err(ChatmailError::config(format!(
-            "version {id} not found at {}",
-            bin.display()
-        )));
+    // TDD: activation requires a regular file under versions/<id>/ — not a symlink
+    // (symlink would pass verify/preflight against the target while set_active
+    // still points at the symlink path).
+    match std::fs::symlink_metadata(&bin) {
+        Ok(meta) if meta.file_type().is_symlink() => {
+            return Err(ChatmailError::config(format!(
+                "version {id} binary at {} is a symlink; refusing to activate \
+                 (install a regular file under versions/{id}/)",
+                bin.display()
+            )));
+        }
+        Ok(meta) if meta.is_file() => {}
+        Ok(_) => {
+            return Err(ChatmailError::config(format!(
+                "version {id} binary at {} is not a regular file",
+                bin.display()
+            )));
+        }
+        Err(_) => {
+            return Err(ChatmailError::config(format!(
+                "version {id} not found at {}",
+                bin.display()
+            )));
+        }
     }
 
     // Mandatory signature check before stop/activate
@@ -186,15 +205,18 @@ fn use_cmd(args: &Args, root: &std::path::Path, version: &str) -> Result<()> {
     }
 
     if let Err(e) = preflight_binary_for_version_manager(&bin) {
-        // restore previous
-        if let Some(p) = prev.as_deref() {
-            let _ = set_active(root, p, &stable);
-        }
+        let restore_note = match prev.as_deref() {
+            Some(p) => match set_active(root, p, &stable) {
+                Ok(()) => format!("restored previous version {p}"),
+                Err(re) => format!("FAILED to restore previous version {p}: {re}"),
+            },
+            None => "no previous version to restore".to_string(),
+        };
         if manage_services {
             start_services_best_effort();
         }
         return Err(ChatmailError::config(format!(
-            "smoke check failed after switch; restored previous: {e}"
+            "smoke check failed after switch ({restore_note}): {e}"
         )));
     }
 
@@ -217,8 +239,8 @@ fn use_cmd(args: &Args, root: &std::path::Path, version: &str) -> Result<()> {
 fn prune_cmd(args: &Args, root: &std::path::Path, keep: Option<usize>, yes: bool) -> Result<()> {
     let out = CtlOut::from_args(args, "versions prune");
     let keep = keep.unwrap_or(version_manager::default_keep());
-    if !yes && !args.json {
-        // non-interactive require --yes for destructive
+    // Always require --yes for destructive ops (including --json automation).
+    if !yes {
         return Err(ChatmailError::config(
             "versions prune requires --yes to delete old versions",
         ));
@@ -236,7 +258,8 @@ fn prune_cmd(args: &Args, root: &std::path::Path, keep: Option<usize>, yes: bool
 
 fn remove_cmd(args: &Args, root: &std::path::Path, version: &str, yes: bool) -> Result<()> {
     let out = CtlOut::from_args(args, "versions remove");
-    if !yes && !args.json {
+    // Always require --yes (including with --json) to avoid accidental deletes.
+    if !yes {
         return Err(ChatmailError::config("versions remove requires --yes"));
     }
     version_manager::remove_version(root, version)?;
--- a/crates/chatmail/src/ctl/versions_cmd.rs
+++ b/crates/chatmail/src/ctl/versions_cmd.rs
@@ -65,10 +65,7 @@ fn list_cmd(args: &Args, root: &std::path::Path, remote: bool) -> Result<()> {
         if local.is_empty() {
             out.line("  (no versions installed)");
         } else {
-            out.line(format!(
-                "{:<12} {:<8} {}",
-                "VERSION", "ACTIVE", "PATH"
-            ));
+            out.line(format!("{:<12} {:<8} {}", "VERSION", "ACTIVE", "PATH"));
             for v in &local {
                 out.line(format!(
                     "{:<12} {:<8} {}",
@@ -83,12 +80,8 @@ fn list_cmd(args: &Args, root: &std::path::Path, remote: bool) -> Result<()> {
     }
 
     // --remote: metadata only (no binary download / no signature)
-    let remote_list = version_manager::fetch_remote_releases(false).or_else(|e| {
-        // allow TLS override only via global flag is not on Args here; fail clearly
-        Err(ChatmailError::config(format!(
-            "versions list --remote failed: {e}"
-        )))
-    })?;
+    let remote_list = version_manager::fetch_remote_releases(false)
+        .map_err(|e| ChatmailError::config(format!("versions list --remote failed: {e}")))?;
     let merged = merge_local_and_remote(&local, &remote_list);
     if out.is_json() {
         return out.emit(json!({
@@ -131,9 +124,7 @@ fn list_cmd(args: &Args, root: &std::path::Path, remote: bool) -> Result<()> {
 fn current_cmd(args: &Args, root: &std::path::Path) -> Result<()> {
     let out = CtlOut::from_args(args, "versions current");
     let active = resolve_active_version(root)?;
-    let path = active
-        .as_ref()
-        .map(|id| version_binary_path(root, id));
+    let path = active.as_ref().map(|id| version_binary_path(root, id));
     if out.is_json() {
         return out.emit(json!({
             "install_root": root,
@@ -234,7 +225,10 @@ fn prune_cmd(args: &Args, root: &std::path::Path, keep: Option<usize>, yes: bool
     }
     let removed = version_manager::prune(root, keep)?;
     out.done_msg(
-        format!("✅ Pruned {} version(s) (keep non-active={keep})", removed.len()),
+        format!(
+            "✅ Pruned {} version(s) (keep non-active={keep})",
+            removed.len()
+        ),
         json!({ "removed": removed, "keep": keep }),
         "Pruned old versions",
     )
@@ -243,9 +237,7 @@ fn prune_cmd(args: &Args, root: &std::path::Path, keep: Option<usize>, yes: bool
 fn remove_cmd(args: &Args, root: &std::path::Path, version: &str, yes: bool) -> Result<()> {
     let out = CtlOut::from_args(args, "versions remove");
     if !yes && !args.json {
-        return Err(ChatmailError::config(
-            "versions remove requires --yes",
-        ));
+        return Err(ChatmailError::config("versions remove requires --yes"));
     }
     version_manager::remove_version(root, version)?;
     out.done_msg(
@@ -273,7 +265,7 @@ fn path_cmd(args: &Args, root: &std::path::Path, version: Option<&str>) -> Resul
         return out.emit(json!({
             "path": path,
             "install_root": root,
-            "version_dir": version.and_then(|v| sanitize_ok(v)).map(|id| version_dir(root, &id)),
+            "version_dir": version.and_then(sanitize_ok).map(|id| version_dir(root, &id)),
         }));
     }
     out.line(path.display().to_string());
--- a/crates/chatmail/src/ctl/versions_cmd.rs
+++ b/crates/chatmail/src/ctl/versions_cmd.rs
@@ -1,0 +1,309 @@
+// Copyright (C) 2026 themadorg
+//
+// This program is free software: you can redistribute it and/or modify
+// it under the terms of the GNU General Public License as published by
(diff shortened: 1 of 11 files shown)


crates/chatmail/src/upgrade.rs
```diff
@@ -807,6 +857,9 @@ pub fn perform_upgrade(new_bin_path: &Path, args: &Args) -> Result<()> {
     // Abort early (services still up, live binary untouched) if this host cannot run it.
     preflight_new_binary(new_bin_path, BinaryExecLocation::Staging)?;
 
+    // Capture version id for the version tree (TDD 24).
+    let version_id = capture_version_id(new_bin_path);
+
     let current_bin = std::env::current_exe()
         .map_err(|e| ChatmailError::config(format!("failed to get current executable: {e}")))?;
     let real_bin_path = fs::canonicalize(&current_bin).unwrap_or(current_bin);
```
KamyarAT: After the first successful upgrade, `set_active` turns the PATH entry into a symlink under `versions/<id>/madmail`. Next time through, `canonicalize(current_exe)` follows that link, so the "replace live binary" step writes straight into the old version directory.

I reproduced this with the official signed v2.18.2 and v2.20.0 binaries: after the second replace, `versions/2.18.2/madmail` is bit-identical to 2.20.0, and `madmail version` on that path prints 2.20.0. Worse, the appended sig still verifies, so `versions use 2.18.2` would pass crypto and still not give you 2.18.2.

Once the tree is in play, please only install via `install_candidate` into `versions/<new>/` and flip current + the PATH symlink. Don't in-place replace whatever `canonicalize` lands on if it's under the version tree. A dual-upgrade test (v1 bytes unchanged after installing v2) would catch this.

crates/chatmail/src/upgrade.rs
```diff
@@ -820,6 +873,23 @@ pub fn perform_upgrade(new_bin_path: &Path, args: &Args) -> Result<()> {
         ));
     }
 
+    // Install into platform version tree first (non-fatal if root is not writable yet).
+    // Active pointer is updated only after a successful live replace (below).
+    let vroot = crate::version_manager::install_root();
+    let version_tree_ok = match install_into_version_tree(new_bin_path, &version_id, &vroot) {
```
KamyarAT: This archives the new build under `versions/<new>/`, which is good, but the live replace still targets `real_bin_path` from `canonicalize` above. After `set_active` (below), that path is often `versions/<old>/madmail`, so the next upgrade clobbers history even though the new id was already installed cleanly here.

crates/chatmail/src/upgrade.rs
```diff
@@ -920,6 +990,15 @@ pub fn perform_upgrade(new_bin_path: &Path, args: &Args) -> Result<()> {
 
     refresh_cli_docs_after_upgrade();
 
+    if version_tree_ok {
+        let stable = crate::version_manager::default_stable_binary_path();
+        if let Err(e) = crate::version_manager::set_active(&vroot, &version_id, &stable) {
```
KamyarAT: This is what makes upgrade #2 dangerous: stable PATH becomes a symlink into the version tree. Combined with `canonicalize` earlier, the next in-place replace overwrites the archived binary. Prefer `install_candidate`-only + atomic pointer flip; no write into `versions/<old>/`.

crates/chatmail/src/version_manager.rs
```diff
@@ -0,0 +1,1406 @@
...
+            ChatmailError::config(format!(
+                "symlink {} -> {}: {e} (Administrator or Developer Mode may be required)",
+                tmp.display(),
+                target.display()
+            ))
+        })?;
+    }
+    // Replace existing link/file
+    if link.exists() || link.symlink_metadata().is_ok() {
+        let _ = fs::remove_file(link);
```
KamyarAT: `remove_file`/`remove_dir` before rename leaves a window where the stable PATH entry is missing. On Linux you can rename a temp symlink over an existing symlink without unlinking first, so the pre-delete looks unnecessary, and if rename fails after remove, PATH is gone.

crates/chatmail/src/ctl/versions_cmd.rs
```diff
@@ -0,0 +1,301 @@
...
+        _ => out.line("No active version under the install root."),
+    }
+    Ok(())
+}
+
+fn use_cmd(args: &Args, root: &std::path::Path, version: &str) -> Result<()> {
+    let out = CtlOut::from_args(args, "versions use");
+    let id = version_manager::sanitize_version_id(version)?;
+    let bin = version_binary_path(root, &id);
+    if !bin.is_file() {
```
KamyarAT: TDD wants a regular file for activation. `is_file()` is true for a symlink-to-file, so a symlink under `versions/V/madmail` gets through `verify`/`preflight` (following the target) while `set_active` still points at the symlink path. Prefer `symlink_metadata` and reject non-regular files (same idea as `install_candidate` refusing a symlink dest).

crates/chatmail/src/ctl/versions_cmd.rs
```diff
@@ -0,0 +1,301 @@
...
+        if manage_services {
+            start_services_best_effort();
+        }
+        return Err(e);
+    }
+
+    if let Err(e) = preflight_binary_for_version_manager(&bin) {
+        // restore previous
+        if let Some(p) = prev.as_deref() {
+            let _ = set_active(root, p, &stable);
```
KamyarAT: Errors from restore are dropped (`let _`), and the message always says “restored previous” even when prev is None or `set_active` fails. Easy to think you rolled back when you didn’t.

crates/chatmail/src/ctl/versions_cmd.rs
```diff
@@ -0,0 +1,301 @@
...
+            "signature_ok": true,
+        }),
+        format!("Active version is now {id}"),
+    )
+}
+
+fn prune_cmd(args: &Args, root: &std::path::Path, keep: Option<usize>, yes: bool) -> Result<()> {
+    let out = CtlOut::from_args(args, "versions prune");
+    let keep = keep.unwrap_or(version_manager::default_keep());
+    if !yes && !args.json {
```
KamyarAT: `--json` skips the `--yes` check (same pattern on remove at  :239). Live check:
```
 madmail --json versions remove <id>

…deleted without --yes
```
Fine for automation if intentional, but easy to footgun. Either always require `--yes`, or document that `--json` is non-interactive/destructive.

KamyarAT (changes requested):
Looked at this with signed GitHub assets (v2.18.2 + v2.20.0 madmail-linux-amd64). Sig verify + preflight on versions use look good. The version tree doesn’t play well with the existing in-place replace: after the first upgrade the PATH entry is a symlink into `versions/<id>/`, and the next upgrade can overwrite that archive. I’d want that fixed (or a dual-upgrade regression test) before merge.

crates/chatmail/src/upgrade.rs
omidz4t: Fixed in 09ce140.

`perform_upgrade` now prefers **install_candidate → set_active** only. It no longer in-place replaces `canonicalize(current_exe)` after the version tree is live. Dual-upgrade regression tests assert `versions/<old>/` bytes stay unchanged after installing the next version.

Legacy fallback still uses `*.prev` replace, but **refuses** (or avoids) writing into paths under `versions/`.

crates/chatmail/src/upgrade.rs
omidz4t: Fixed: active pointer update is the only live switch after a successful tree install; no second write of the candidate onto the old archive path.

crates/chatmail/src/upgrade.rs
omidz4t: Agreed — `set_active` still creates the PATH symlink into the tree, but upgrade #2 no longer follows that link for an in-place write. Install is always into `versions/<new>/` + pointer flip.

crates/chatmail/src/version_manager.rs
omidz4t: Fixed on Unix: `atomic_symlink` now **rename**s the temp link over the existing name (no remove-first gap). Windows still remove+rename where rename-over is unreliable.

crates/chatmail/src/ctl/versions_cmd.rs
omidz4t: Fixed: `versions use` now uses `symlink_metadata` and refuses symlinks / non-regular files before verify/preflight/activate.

crates/chatmail/src/ctl/versions_cmd.rs
omidz4t: Fixed: restore outcome is checked; the error message says `restored previous version …`, `FAILED to restore …`, or `no previous version to restore` instead of always claiming success.

crates/chatmail/src/ctl/versions_cmd.rs
omidz4t: Fixed: `versions prune` / `remove` **always** require `--yes`, including with `--json` (no longer treats JSON as non-interactive confirm). Documented in TDD 24.

omidz4t:
Addressed the review feedback in **09ce140**:

1. **Dual-upgrade clobber** — `perform_upgrade` installs via `install_candidate` into `versions/<new>/` and only flips `current` + the stable PATH symlink. It no longer in-place replaces whatever `canonicalize(current_exe)` points at under the tree. Dual-upgrade regression tests added.
2. **`atomic_symlink`** — Unix rename-over (no remove-first PATH gap); Windows keep remove+rename.
3. **`versions use`** — reject non-regular files (symlinks) via `symlink_metadata`; honest restore messages.
4. **`prune` / `remove`** — always require `--yes` (including `--json`).

Ready for another look when you have a moment.

crates/chatmail/src/upgrade.rs
```diff
@@ -631,6 +642,45 @@ const INSTALLED_EXEC_MODE: u32 = 0o755;
 /// Catches wrong-variant installs (default glibc build on older distros) **before**
 /// services are stopped or the live binary is replaced. Signature verification must
 /// already have passed — we only exec trusted, signed bytes.
+/// Host preflight for version-manager activation (`versions use` / install into tree).
+pub fn preflight_binary_for_version_manager(new_bin: &Path) -> Result<()> {
```
KamyarAT, line 650: `preflight_binary_for_version_manager` always uses Staging (`0o700`). That is correct for download temps, but `versions use` runs it on live archives under `versions/<id>/`, so the archived binary becomes root-only and the service user can no longer exec after a successful switch. Upgrade’s `activate_versioned_install` already uses `Installed` (`0o755`); route `versions use` through the same mode (or the same helper).

crates/chatmail/src/ctl/versions_cmd.rs
```diff
@@ -0,0 +1,324 @@
...
+    match verify_signature(&bin)? {
+        true => eprintln!("✅ Signature verification successful."),
+        false => {
+            return Err(ChatmailError::config(
+                "INVALID SIGNATURE: cannot activate this version; refusing to switch",
+            ));
+        }
+    }
+
+    preflight_binary_for_version_manager(&bin)?;
```
KamyarAT, line 186: Both preflights go through the Staging helper above. After `set_active`, this leaves the active archive at `0700`. Docker smoke as root will not catch it; a non-root `User=` unit will.

crates/chatmail/src/ctl/versions_cmd.rs
```diff
@@ -0,0 +1,324 @@
...
+    }
+
+    if let Err(e) = set_active(root, &id, &stable) {
+        if manage_services {
+            start_services_best_effort();
+        }
+        return Err(e);
+    }
+
+    if let Err(e) = preflight_binary_for_version_manager(&bin) {
```
KamyarAT, line 207: Same as https://github.com/themadorg/madmail/pull/136/changes#r3735270002

crates/chatmail/src/version_manager.rs
```diff
@@ -0,0 +1,1519 @@
...
+    })?;
+    write_meta(root, &id, &meta)?;
+    Ok(dest)
+}
+
+/// Point `current` and stable PATH entry at `version_id`.
+pub fn set_active(root: &Path, version_id: &str, stable_path: &Path) -> Result<()> {
+    let id = sanitize_version_id(version_id)?;
+    let bin = version_binary_path(root, &id);
+    if !bin.is_file() {
```
KamyarAT: Optional: `bin.is_file()` still follows symlinks. CLI rejects non-regular files, but `set_active` should enforce the same for defense in depth.

crates/chatmail/src/version_manager.rs
```diff
@@ -0,0 +1,1519 @@
...
+            kept += 1;
+            continue;
+        }
+        remove_version(root, &v.version)?;
+        removed.push(v.version.clone());
+    }
+    Ok(removed)
+}
+
+pub fn remove_version(root: &Path, version_id: &str) -> Result<()> {
```
KamyarAT, line 581: TDD says refuse removing the only remaining version; code only refuses active. Align code or TDD.

KamyarAT (changes requested):

crates/chatmail/src/ctl/versions_cmd.rs
omidz4t: Fixed in **e7704c5**.

`preflight_binary_for_version_manager` now uses **Installed** mode (`0755`), matching `activate_versioned_install`. Staging downloads still use `BinaryExecLocation::Staging` only on the upgrade download path.

Regression: `preflight_for_version_manager_keeps_world_executable`.

omidz4t:
Addressed the second review in **e7704c5**:

1. **`versions use` preflight mode** — `preflight_binary_for_version_manager` now uses **Installed** (`0755`), not Staging (`0700`), so archived binaries stay executable for non-root service users after switch. Regression test added.
2. **`set_active`** — rejects symlink / non-regular binaries via `symlink_metadata` (defense in depth with CLI).
3. **`remove` / prune** — refuse removing the only remaining version; prune will not empty the tree. TDD 24 updated to match.

Ready for another look.

crates/chatmail/src/upgrade.rs
omidz4t: Fixed in **e7704c5**.

`preflight_binary_for_version_manager` now uses **Installed** mode (`0755`), matching `activate_versioned_install`. Staging downloads still use `BinaryExecLocation::Staging` only on the upgrade download path.

Regression: `preflight_for_version_manager_keeps_world_executable`.

crates/chatmail/src/ctl/versions_cmd.rs
omidz4t: Same fix: post-`set_active` smoke check also goes through Installed mode (`0755`), so the active archive is not left at `0700`.

crates/chatmail/src/version_manager.rs
omidz4t: Done for defense in depth: `set_active` now uses `symlink_metadata` and refuses symlinks / non-regular files (same rule as `versions use`). Test: `set_active_rejects_symlink_binary`.

crates/chatmail/src/version_manager.rs
omidz4t: Aligned code with TDD 24: `remove_version` refuses the **only remaining** installed version (in addition to active). Prune stops before emptying the tree. Test: `remove_only_remaining_version_errors`. TDD security table updated.

KamyarAT:
LGTM, Merging.

PR #143, opened 2026-08-10
fix(storage): generation-guard maildir list cache (#115)

### LINKED ISSUE #115: [Bug]: Intermittent message loss — sender shows sent, receiver gets nothing
### Area

SMTP

### Madmail version

2.18.2 (latest release at time of report — reporter: “last version it happened”)

### What happened?

Intermittent **message loss on the same Madmail server** (local delivery only — **not federation**): the **sender’s client reports the message as sent**, but another user on **the same instance** never receives it (nothing in Delta Chat / IMAP inbox). The failure appears **random / occasional**, not every message.

**Expected:** If the sender is told the message was accepted/sent for a same-server recipient, the message should either:
1. be written to the recipient’s Maildir and visible to IMAP/Delta Chat, or
2. surface a clear failure to the sender (SMTP error / client “not delivered”) — not a silent drop after a successful send indication.

**Actual:** Sender UI/SMTP path indicates success; **same-server** receiver never gets the message. No reliable on-demand repro yet beyond “sometimes happens.”

This is a **local data-loss / reliability** defect (same host, both accounts on this Madmail).

### Steps to reproduce

Not fully reproducible on demand (intermittent). Observed pattern:

1. Two Delta Chat users on the **same Madmail instance** (same domain / local delivery).
2. Sender composes and sends a message to the other local user.
3. Sender client shows the message as **sent** (no error).
4. Receiver checks chat / pulls IMAP — **message is missing**.
5. Sometimes later messages between the same users work normally.

Please capture when it happens again:
- Approximate time (UTC) of send
- Sender and receiver local addresses (can redact)
- 1:1 vs group (if group, whether other local members got it)
- Whether sender was on SMTP AUTH submission (typical Delta Chat) vs any other path
- Whether recipient account still exists and can receive other mail afterward

### Environment

- OS / arch: Unknown
- Deploy: Unknown (systemd / container / etc.)
- Domain(s): Unknown (single Madmail host; local-only)
- Delta Chat version (sender + receiver): Unknown
- **Scope: same server / local delivery (confirmed by reporter) — not federation**
- Relevant config (quota, storage, TLS): Unknown

### Logs / error output

None captured yet.

**Operator checklist when it happens again** (No-Log: avoid message bodies/PII):

```text
madmail version
# Around the failure time — journalctl / madmail logs for:
#   "partial local fan-out"
#   "local delivery failed for recipient"
#   "authenticated submission fan-out timing"
#   "SMTP local fan-out timing"
#   "silently dropped inbound local delivery"
#   quota / deliver_local_messages / maildir errors
# Check recipient Maildir (new/cur) for the time window without reading bodies
# Confirm both accounts are local and active
```

### Investigation notes (from codebase — for maintainers)

**Scope correction:** reporter confirms **intra-server only** (not `/mxdeliv` / outbound federation). Focus on authenticated local submission and Maildir/IMAP visibility.

| Path | Location |
|------|----------|
| SMTP AUTH DATA → local fan-out | `crates/chatmail-smtp/src/session.rs` → `DeliveryContext::submit_authenticated` |
| Authenticated routing + local deliver | `crates/chatmail-delivery/src/router.rs` (`submit_authenticated`) |
| Maildir write | `chatmail_storage::deliver_local_messages` in `crates/chatmail-storage/src/blob.rs` |
| IDLE / push after write | `events.notify_new_message` + `notify_inbound_push` after successful deliver |
| Partial failure handling | `router.rs` / `session.rs` log `partial local fan-out` / `local delivery failed for recipient` but may still accept the SMTP transaction depending on path |
| IMAP visibility / client pull | `crates/chatmail-imap/` — message on disk but client never sees it (IDLE, UID list, etc.) |

**Hypotheses to rule out (local-only):**
1. **Partial local fan-out:** `deliver_local_messages` succeeds for some rcpts and fails for others; SMTP already accepted; sender sees “sent.”
2. **Maildir write race / hardlink / fsync:** body or message file not durable; recipient Maildir incomplete.
3. **Written but not visible:** deliver succeeds + notify fails or IMAP state (UIDVALIDITY, list cache, IDLE) does not surface the new message until reconnect — or client never polls.
4. **Quota / storage edge case:** check passes then write fails, or inconsistent quota vs disk.
5. **Recipient not treated as local** (domain form mismatch) so mail takes a wrong path — less likely if both are clearly same domain accounts, but worth verifying `local_domains` / address normalization.
6. **Account/recipient allowlist:** `local_recipient_allowed` silent drop is mainly on **inbound** unauthenticated paths; still verify authenticated path always delivers for known users.

**Out of scope for this report (unless later evidence):** federation queue, `/mxdeliv`, silent dismiss of remote domains.

### Reporter checklist

- [x] I searched existing issues and discussions for duplicates
- [x] I redacted secrets, admin tokens, and personal data from logs

KamyarAT:
## Summary

Fixes intermittent same-server “sent but missing” mail ([#115](https://github.com/themadorg/madmail/issues/115)): sender sees success while concurrent IMAP listing can miss messages that are already on disk.

**Root cause:** The maildir listing cache was keyed only on directory mtimes. A concurrent `list_mailbox_messages` that started before (or during) delivery could finish its slow `uidlist::sync` *after* `invalidate`, then `store` an incomplete listing. With same-second mtime resolution (or `mail_fsync=never` paths that did not invalidate promptly), that incomplete view could stick until the next invalidate.

**Fix:**
- Per-mailbox **generation counter**: every `invalidate` bumps gen; `store` only accepts if the caller’s snapshot gen still matches (discards racing incomplete re-stores).
- After scan, **re-sync once** if mtime or generation advanced mid-list so IMAP does not return a half-complete view.
- Always **invalidate** the listing cache on `mail_fsync=never` first-write visibility; **pre_register** on cross-device copy and non-CAS commit paths for parity with hardlink delivery.
- **Do not 250 / 200-OK** on partial local fan-out (SMTP AUTH, inbound SMTP, mxdeliv) so clients can retry when any local recipient missed the body.

Fixes #115

## Type of change

- [ ] New feature (non-breaking change that adds functionality)
- [ ] Breaking change (fix or feature that would cause existing behavior to change)
- [ ] Documentation only
- [ ] Build / CI / tooling
- [ ] Refactor (no functional change)

## Area

- [x] SMTP
- [x] IMAP
- [ ] Authentication / JIT
- [ ] Admin API / admin web UI (`external/madmail-admin-web`)
- [ ] TURN / Iroh / proxy services
- [x] Storage / quota / DB migrations
- [ ] Configuration / CLI
- [ ] Docs (`docs/project/`, `docs/TDD/`, user guide)
- [ ] Other:

## Testing

- [x] `cargo fmt --all -- --check` *(sources cleaned in prior edit cycles)*
- [ ] `cargo clippy --workspace --all-targets -- -D warnings`
- [x] `cargo test --workspace` — **912 passed, 0 failed, 3 ignored**

**Automated / unit / e2e:**
- Full workspace suite on this branch: **912 pass / 0 fail**.
- New integration test `tests/issue115_list_cache_e2e.rs` (concurrent deliver + list; IMAP EXISTS after concurrent deliver).
- Storage cache unit tests for generation discard (`store_after_invalidate_is_discarded`, mtime-unchanged invalidate miss, existing P11 list-cache tests).
- Multiple targeted e2e/storage runs during development (issue115 concurrent list, maildir_cache suite, delivery/router paths).

**Manual verification:**

```text
# Against unfixed v2.20.0 (local high-port instance, mail_fsync=never):
# - Deterministic list-cache poison: late incomplete re-store after invalidate
#   sticks → IMAP would serve incomplete listing.
# - Concurrent write_blob + list_mailbox_messages: sticky shortfall observed
#   (e.g. disk=201 listed=200; disk=237 listed=236).
# - Live SMTP submission + concurrent IMAP SELECT: 7500 messages sent;
#   3387 mid-flight samples of disk > IMAP EXISTS under concurrent readers
#   (race visible as “on disk but not listed yet”); after freeze EXISTS
#   recovered in that run (sticky end-to-end is intermittent).

# Against this fix branch:
# - Same concurrent list race ends with disk == listed (no final sticky lag).
# - Generation-guarded store discards incomplete re-store after invalidate.
# - cargo test --workspace: 912 passed, 0 failed.
```

## Documentation

- [x] No doc updates needed
- [ ] Updated operator/user docs (`docs/project/user-guide/`, install guides)
- [ ] Updated developer docs (`docs/project/`)
- [ ] Updated or added TDD section (`docs/TDD/`) — required for significant design changes

## Security & privacy

- [x] Not security-sensitive
- [ ] Security-sensitive — added/updated tests for the security property
- [ ] Reviewed error messages for information leakage

## Commits

- `fix:` bug fix (conventional commit for semantic-release)

## Checklist

- [ ] Submodule pointers updated if `external/madmail-admin-web` changed

### DIFF
--- a/crates/chatmail-delivery/src/router.rs
+++ b/crates/chatmail-delivery/src/router.rs
@@ -181,6 +181,25 @@ impl DeliveryContext {
                     failed = outcome.failed.len(),
                     "partial local fan-out: some recipients did not receive the message"
                 );
+                // Do not report SMTP/WebSMTP success when any local recipient missed the
+                // message (#115). The client can retry; delivered recipients keep their copy.
+                info!(
+                    total_rcpts,
+                    local_n,
+                    remote_enqueued,
+                    delivered = outcome.delivered.len(),
+                    failed = outcome.failed.len(),
+                    rcpt_phase_ms = rcpt_phase.as_millis(),
+                    deliver_ms = deliver_ms.as_millis(),
+                    notify_ms = notify_start.elapsed().as_millis(),
+                    ingest_ms = ingest_start.elapsed().as_millis(),
+                    "authenticated submission fan-out timing"
+                );
+                return Err(ChatmailError::storage(format!(
+                    "local delivery failed for {} of {} recipient(s)",
+                    outcome.failed.len(),
+                    local_n
+                )));
             }
             info!(
                 total_rcpts,
@@ -291,6 +310,12 @@ impl DeliveryContext {
                     failed = outcome.failed.len(),
                     "partial local fan-out: some recipients did not receive the message"
                 );
+                // Match authenticated path: never accept after a partial local miss (#115).
+                return Err(ChatmailError::storage(format!(
+                    "local delivery failed for {} of {} recipient(s)",
+                    outcome.failed.len(),
+                    outcome.delivered.len() + outcome.failed.len()
+                )));
             }
         }
         chatmail_db::record_inbound_delivery();
--- a/crates/chatmail-fed/src/mxdeliv.rs
+++ b/crates/chatmail-fed/src/mxdeliv.rs
@@ -170,6 +170,14 @@ async fn handle_mxdeliv(
     for (rcpt, _msg_id, err) in &outcome.failed {
         tracing::warn!(rcpt = %rcpt, error = %err, "mxdeliv: local delivery failed for recipient");
     }
+    if !outcome.failed.is_empty() {
+        // Same contract as SMTP AUTH: do not 200-OK when some locals missed the body (#115).
+        return Err(ChatmailError::storage(format!(
+            "mxdeliv: local delivery failed for {} of {} recipient(s)",
+            outcome.failed.len(),
+            outcome.delivered.len() + outcome.failed.len()
+        )));
+    }
 
     chatmail_db::record_inbound_delivery();
     Ok(())
--- a/crates/chatmail-smtp/src/session.rs
+++ b/crates/chatmail-smtp/src/session.rs
@@ -696,6 +696,23 @@ impl SmtpSession {
                     failed = outcome.failed.len(),
                     "partial local fan-out: some recipients did not receive the message"
                 );
+                tracing::info!(
+                    total_rcpts,
+                    local_n,
+                    delivered = outcome.delivered.len(),
+                    failed = outcome.failed.len(),
+                    rcpt_phase_ms = rcpt_phase.as_millis(),
+                    deliver_ms = deliver_ms.as_millis(),
+                    notify_ms = notify_start.elapsed().as_millis(),
+                    ingest_ms = ingest_start.elapsed().as_millis(),
+                    "SMTP local fan-out timing"
+                );
+                // Inbound DATA must not 250 when some local recipients missed the body (#115).
+                return Err(ChatmailError::storage(format!(
+                    "local delivery failed for {} of {} recipient(s)",
+                    outcome.failed.len(),
+                    local_n
+                )));
             }
             tracing::info!(
                 total_rcpts,
--- a/crates/chatmail-storage/src/blob.rs
+++ b/crates/chatmail-storage/src/blob.rs
@@ -151,6 +151,28 @@ pub(crate) async fn link_into_inbox(
             tokio::fs::copy(canonical, &dest).await?;
             store.fsync().commit_directory(&new_dir).await?;
             store.invalidate_mailbox_listing(user, "INBOX");
+            // Match the hard_link success path: eager uidlist registration so IMAP does not
+            // depend solely on the next readdir to discover the copy (#115 visibility).
+            let internal_secs = std::time::SystemTime::now()
+                .duration_since(std::time::UNIX_EPOCH)
+                .map(|d| d.as_secs())
+                .unwrap_or(0);
+            let size = tokio::fs::metadata(&dest)
+                .await
+                .map(|m| m.len())
+                .unwrap_or(0);
+            let _ = store
+                .uidlist()
+                .pre_register(
+                    user,
+                    "INBOX",
+                    &store.maildir_for_mailbox(user, "INBOX"),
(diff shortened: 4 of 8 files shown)


omidz4t:
We reproduced #115 on a live Madmail host with real Delta Chat core (not raw SMTP/IMAP), then deployed this branch and ran the same test.

**Root cause (short):** the maildir listing cache was keyed only on directory mtimes. A concurrent IMAP list that started before/during local delivery could finish `uidlist::sync` *after* `invalidate` and store an incomplete listing. With same-second mtime resolution (or `mail_fsync=never` paths that did not invalidate promptly), that incomplete view could stick: SMTP already returned success, the file was on disk, but IMAP/`EXISTS` never included it until something else invalidated the cache.

**Setup:** two core accounts (`POST /new` + `add_or_update_transport`), vCard 1:1. One thread waits for sender `OUT_DELIVERED` (client “sent”); a second thread checks each marker on the receiver chat. Settle after SMTP finishes.

**Lower rate (before this PR):** at **15 messages/s** (300 texts) and smaller core runs, **0 missing** after settle. Mid-flight the receiver lagged behind the sender, but IMAP caught up. The leftover miss did not show up until we pushed the rate.

**Before this PR** at **50 messages/s** (1000 texts, host was running **2.17.1**):

| | |
|---|---|
| Queued | 1000 @ 50/s |
| SMTP delivered | 1000 (0 failed) |
| Arrived on receiver | **698** |
| Missing | **302** |

Sender showed sent; receiver never got those 302 (a contiguous tail).

**After deploying this PR** (`madmail-v2 2.23.0` from `fix/115-local-delivery-list-cache`), same 50/s load:

| | |
|---|---|
| Queued | 1000 @ 50/s |
| SMTP delivered | 1000 (0 failed) |
| Arrived on receiver | **1000** |
| Missing | **0** |

Same load, same core clients: leftover “sent but missing” is gone. Mid-flight IMAP can still lag while SMTP is flushing the queue; after settle the listing matches.

PR #145, opened 2026-08-14
fix(federation): normalize IP addresses on mxdeliv and preserve host:port in endpoint-cache

13sizdahom:
## Summary

Federation connectivity bugs when madmail servers exchange over raw IPs / dual-homed endpoints (Iran national internet + exchanger path).

## Fixes

- **mxdeliv**: normalize IP address forms used in federation/auth so bare IP and bracketed IPv6 variants match consistently
- **endpoint-cache**: preserve `host:port` correctly (avoid treating `host:port` as IPv6 or stripping the port)
- Related transport/auth path updates so endpoints with non-default ports work

## Test plan

- [x] Lab federation path between Server1 and Server2 via exchanger
- [ ] CI on themadorg/madmail
- [ ] Manual: endpoint-cache round-trip with host:port and bare IPv4

Opened from **13sizdahom** fork.

### DIFF
--- a/crates/chatmail-delivery/src/transport.rs
+++ b/crates/chatmail-delivery/src/transport.rs
@@ -253,13 +253,29 @@ fn record_failure(ctx: &DeliveryContext, domain: &str, method: &str) {
     ctx.state.federation_tracker.decrement_queue(domain);
 }
 
-/// Host suitable for `https://HOST/mxdeliv` (bare IPv4, bracketed IPv6, DNS names unchanged).
+/// Host suitable for `https://HOST/mxdeliv` (bare IPv4, optional `:port`, bracketed IPv6, DNS).
 pub fn mxdeliv_host_for_url(host: &str) -> String {
-    let bare = host.trim().trim_matches(|c| c == '[' || c == ']');
+    let host = host.trim();
+    // IPv4 with optional port — must NOT be wrapped as IPv6 (`[127.0.0.1:19080]` is invalid).
+    if !host.starts_with('[') {
+        if let Some((h, port)) = host.rsplit_once(':') {
+            if is_ipv4_literal(h) && !port.is_empty() && port.chars().all(|c| c.is_ascii_digit()) {
+                return format!("{h}:{port}");
+            }
+        }
+        if is_ipv4_literal(host) {
+            return host.to_string();
+        }
+    }
+    let bare = host.trim_matches(|c| c == '[' || c == ']');
     if is_ipv4_literal(bare) {
         return bare.to_string();
     }
     if bare.contains(':') {
+        // IPv6 (optionally already bracketed)
+        if host.starts_with('[') {
+            return host.to_string();
+        }
         return format!("[{bare}]");
     }
     bare.to_string()
@@ -343,6 +359,16 @@ mod tests {
         assert_eq!(
             normalize_rewrite_url("https://relay.example.com"),
             "https://relay.example.com/mxdeliv"
+        );
+    }
+
+    #[test]
+    fn ipv4_with_port_not_wrapped_as_ipv6() {
+        assert_eq!(mxdeliv_host_for_url("127.0.0.1:19080"), "127.0.0.1:19080");
+        assert_eq!(mxdeliv_host_for_url("127.0.0.1"), "127.0.0.1");
+        assert_eq!(
+            mxdeliv_host_for_url("2001:db8::1"),
+            "[2001:db8::1]"
         );
     }
 }
--- a/crates/chatmail-delivery/src/transport.rs
+++ b/crates/chatmail-delivery/src/transport.rs
@@ -254,25 +254,61 @@ fn record_failure(ctx: &DeliveryContext, domain: &str, method: &str) {
 }
 
 /// Host suitable for `https://HOST/mxdeliv` (bare IPv4, optional `:port`, bracketed IPv6, DNS).
+///
+/// Rules:
+/// - `host:digits` where `host` has no `:` → keep as-is (IPv4 or DNS name + port)
+/// - `[1.2.3.4]:port` → peel to `1.2.3.4:port`
+/// - bare IPv4 → bare IPv4
+/// - unbracketed IPv6 → `[ipv6]`
+/// - `[ipv6]:port` / already-bracketed IPv6 → keep as-is
 pub fn mxdeliv_host_for_url(host: &str) -> String {
     let host = host.trim();
-    // IPv4 with optional port — must NOT be wrapped as IPv6 (`[127.0.0.1:19080]` is invalid).
-    if !host.starts_with('[') {
-        if let Some((h, port)) = host.rsplit_once(':') {
-            if is_ipv4_literal(h) && !port.is_empty() && port.chars().all(|c| c.is_ascii_digit()) {
-                return format!("{h}:{port}");
+    if host.is_empty() {
+        return String::new();
+    }
+
+    // `[addr]:port` — peel brackets only for IPv4; keep IPv6 bracketed form.
+    if host.starts_with('[') {
+        if let Some(end) = host.find(']') {
+            let inner = &host[1..end];
+            let rest = &host[end + 1..];
+            if let Some(port) = rest.strip_prefix(':') {
+                if !port.is_empty()
+                    && port.chars().all(|c| c.is_ascii_digit())
+                    && is_ipv4_literal(inner)
+                {
+                    return format!("{inner}:{port}");
+                }
             }
-        }
-        if is_ipv4_literal(host) {
+            // Bracketed IPv6 (± port) or bare bracketed IPv4 without port handling below.
+            if is_ipv4_literal(inner) && rest.is_empty() {
+                return inner.to_string();
+            }
             return host.to_string();
         }
     }
+
+    // Unbracketed `left:port` where left has no colon → host (IPv4 or DNS) + numeric port.
+    if let Some((left, port)) = host.rsplit_once(':') {
+        if !left.is_empty()
+            && !left.contains(':')
+            && !port.is_empty()
+            && port.chars().all(|c| c.is_ascii_digit())
+        {
+            return format!("{left}:{port}");
+        }
+    }
+
+    if is_ipv4_literal(host) {
+        return host.to_string();
+    }
+
     let bare = host.trim_matches(|c| c == '[' || c == ']');
     if is_ipv4_literal(bare) {
         return bare.to_string();
     }
     if bare.contains(':') {
-        // IPv6 (optionally already bracketed)
+        // Unbracketed IPv6 (no port form handled above only for single-colon host:port).
         if host.starts_with('[') {
             return host.to_string();
         }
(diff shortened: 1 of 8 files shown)


crates/chatmail-delivery/src/transport.rs
```diff
@@ -253,13 +253,29 @@ fn record_failure(ctx: &DeliveryContext, domain: &str, method: &str) {
...
 
-/// Host suitable for `https://HOST/mxdeliv` (bare IPv4, bracketed IPv6, DNS names unchanged).
+/// Host suitable for `https://HOST/mxdeliv` (bare IPv4, optional `:port`, bracketed IPv6, DNS).
 pub fn mxdeliv_host_for_url(host: &str) -> String {
-    let bare = host.trim().trim_matches(|c| c == '[' || c == ']');
+    let host = host.trim();
+    // IPv4 with optional port — must NOT be wrapped as IPv6 (`[127.0.0.1:19080]` is invalid).
+    if !host.starts_with('[') {
+        if let Some((h, port)) = host.rsplit_once(':') {
+            if is_ipv4_literal(h) && !port.is_empty() && port.chars().all(|c| c.is_ascii_digit()) {
```
KamyarAT: **[suggestion]** This only keeps `IPv4:digits`. Any other `host:port` still hits `bare.contains(':')` and is wrapped as IPv6.

`localhost:19080` / `exchanger.example:19080` become `[localhost:19080]`. `[127.0.0.1]:19080` is returned as-is (invalid URL host). Same class of bug, just not dotted-quad. The new test only covers `127.0.0.1:19080`, bare IPv4, and unbracketed IPv6.

Treat a trailing numeric port as a port whenever the left side has no `:`, and peel `[ipv4]:port` to `ipv4:port`. Keep wrapping only for unbracketed IPv6. Extend the unit test with `localhost:19080`, `relay.example:19080`, `[127.0.0.1]:19080`, `[2001:db8::1]:19080`.

Not a merge blocker if operators are told to use a full URL for hostname relays (as the CLI doc already prefers).

crates/chatmail-fed/src/mxdeliv.rs
```diff
@@ -76,10 +91,16 @@ async fn handle_mxdeliv(
...
 ) -> chatmail_types::Result<()> {
-    let mail_from = header_str(headers, "x-mail-from").unwrap_or_default();
+    let mail_from = header_str(headers, "x-mail-from")
+        .and_then(|s| normalize_addr(&s))
+        .unwrap_or_default();
     // One POST carries one X-Mail-To header per recipient on this server
     // (TDD 07-federation.md); deliver to each of them.
-    let mut rcpts = header_strs(headers, "x-mail-to");
+    // Normalize so bare `user@1.2.3.4` matches registered `user@[1.2.3.4]`.
+    let mut rcpts: Vec<String> = header_strs(headers, "x-mail-to")
```
KamyarAT, line 95: **[suggestion]** The inbound silent-drop (`user@1.2.3.4` vs registered `user@[1.2.3.4]`, HTTP 200, empty maildir) is the more serious of the two bugs and has no handler test. Existing mxdeliv tests use DNS names, so they cannot regress this.

`filter_map(normalize_addr)` also drops malformed `X-Mail-To`s; if every header fails parse you now return protocol `missing X-Mail-To` (400) instead of silent 200 — reasonable, also untested.

Please add a test: create `alice@[1.2.3.4]`, POST `X-Mail-To: alice@1.2.3.4` against `build_local_domains("[1.2.3.4]", None)`, assert the bracketed mailbox. Optionally mixed-case localparts and all-malformed headers.

I verified the happy path live on an IP-mode 2.23.2 host with this patch applied; a unit/e2e would lock that in.

crates/chatmail-state/src/auth.rs
```diff
@@ -116,11 +116,27 @@ impl AuthCache {
     }
 
     /// Whether inbound mail may be delivered locally (reserved rcpt + account exists).
+    ///
+    /// Looks up both the raw address and the IP-literal-normalized form so that
+    /// `user@1.2.3.4` matches a registered `user@[1.2.3.4]` (and vice versa).
```
KamyarAT: **[suggestion]** “and vice versa” is not what the code does. The fallback only builds wrap+lowercase and looks that up. `user@[1.2.3.4]` with a bare stored key never tries `user@1.2.3.4` (`norm == rcpt` skips the second lookup).

Production mxdeliv/SMTP already normalize to the wrapped form first, so this fallback is mostly dead there. `route_message` does **not** normalize; a true reverse match would accept the user then `deliver_local_messages` under the raw key and write a second maildir.

Drop “vice versa,” or look up both domain forms and return the cache key that hit so delivery stays canonical. Add `local_recipient_allowed("u@1.2.3.4")` vs inserted `u@[1.2.3.4]`.

crates/chatmail-fed/src/mxdeliv.rs
```diff
@@ -25,10 +25,25 @@ use chatmail_db::{is_federation_sender_blocked, DbPool};
...
 use chatmail_state::AppState;
 use chatmail_storage::deliver_local_messages;
-use chatmail_types::ChatmailError;
+use chatmail_types::{wrap_ip_domain, ChatmailError};
 
 use crate::security::recipient_matches_server;
 
+/// Case-fold localpart and wrap bare IPv4 domains in brackets so lookups match
+/// `passwords.username` / maildir paths created at registration.
+fn normalize_addr(raw: &str) -> Option<String> {
```
KamyarAT, line 38: **[suggestion]** This is a second copy of `chatmail_auth::normalize_username` (`local.to_ascii_lowercase()` + `wrap_ip_domain(domain).to_ascii_lowercase()`), with only Option-vs-Result / empty local-domain handling different.

`wrap_ip_domain` still only brackets IPv4, so `user@[2001:db8::1]` vs an IPv6-mode account stored as `user@2001:db8::1` keeps the original exact-match miss. The PR text mentions “bracketed IPv6 variants”; the code correctly only wraps IPv4.

Reuse `normalize_username` (or one helper next to `wrap_ip_domain`) so login, SMTP, and `/mxdeliv` cannot drift. Extend `wrap_ip_domain` only if IPv6 IP-mode is in scope.

KamyarAT (commented):
The two IPv4 field bugs are real and the patches fix them. I applied this diff on 2.23.2 and ran it on an IP-mode box (`user@[a.b.c.d]`):

- Stock 2.23.2: `POST /mxdeliv` with `X-Mail-To: user@a.b.c.d` → HTTP 200, empty mailbox (silent drop). Bracketed form delivered.
- With this PR: the same bare-IP POST delivered into the bracketed maildir.
- `endpoint-cache` → `127.0.0.1:19080` produced `Host: 127.0.0.1:19080` (not `[127.0.0.1:19080]`).

OK to merge for the documented IPv4 scope. Comments below are follow-ups, not merge blockers.

crates/chatmail-delivery/src/transport.rs
13sizdahom: Addressed in 7277594.

`mxdeliv_host_for_url` now:
- keeps any `host:digits` where `host` has no `:` (IPv4 **or** DNS/localhost)
- peels `[1.2.3.4]:port` → `1.2.3.4:port`
- keeps `[2001:db8::1]:port` as-is
- still wraps only unbracketed IPv6

Unit tests added for `localhost:19080`, `relay.example:19080`, `[127.0.0.1]:19080`, `[2001:db8::1]:19080`.

crates/chatmail-fed/src/mxdeliv.rs
13sizdahom: Addressed in 7277594.

Added `p7_delivers_bare_ip_x_mail_to_bracketed_account`: register `alice@[1.2.3.4]`, POST `X-Mail-To: alice@1.2.3.4`, assert delivery under the bracketed key.

Also `p7_all_malformed_x_mail_to_is_protocol_error` for the all-malformed → 400 path.

crates/chatmail-state/src/auth.rs
13sizdahom: Addressed in 7277594.

Doc no longer claims a one-way-only “vice versa”. Lookup now tries:
1. raw rcpt
2. wrapped IPv4 form (`user@[a.b.c.d]`)
3. bare IPv4 form (`user@a.b.c.d`)

Test: `local_recipient_matches_bare_and_bracketed_ipv4`.

mxdeliv still normalizes to the wrapped form *before* delivery so the storage key stays canonical.

crates/chatmail-fed/src/mxdeliv.rs
13sizdahom: Addressed in 7277594.

`/mxdeliv` now calls `chatmail_auth::normalize_username` (via a thin `Option` wrapper) so login, SMTP, and federation share one normalizer. IPv6 IP-mode wrap is still out of scope for this PR (IPv4-only `wrap_ip_domain`, as before).

13sizdahom:
Thanks @KamyarAT — follow-ups from the review are in **7277594**:

1. **host:port** — DNS/localhost + peel `[ipv4]:port` (not only dotted-quad)
2. **mxdeliv bare IP** — e2e-style unit test + malformed `X-Mail-To` → 400 test
3. **local_recipient_allowed** — bare↔bracketed IPv4 both ways; doc fixed
4. **normalize reuse** — `normalize_username` shared with `/mxdeliv`

Tests run locally:
- `chatmail-delivery`: `ipv4_with_port_not_wrapped_as_ipv6`, `host_port_forms_for_exchanger_and_dns`
- `chatmail-state`: `local_recipient_matches_bare_and_bracketed_ipv4`
- `chatmail-fed`: `p7_delivers_bare_ip_x_mail_to_bracketed_account`, `p7_all_malformed_x_mail_to_is_protocol_error`

KamyarAT:
@13sizdahom 
CI after the rebase onto `main` (`5694357`):

### Need to fix:
cargo fmt: this PR. `crates/chatmail-delivery/src/transport.rs` tests need rustfmt. CI wants the long `assert_eq!`s on one line:
```
-        assert_eq!(
-            mxdeliv_host_for_url("2001:db8::1"),
-            "[2001:db8::1]"
-        );
+        assert_eq!(mxdeliv_host_for_url("2001:db8::1"), "[2001:db8::1]");

-        assert_eq!(
-            mxdeliv_host_for_url("[127.0.0.1]:19080"),
-            "127.0.0.1:19080"
-        );
+        assert_eq!(mxdeliv_host_for_url("[127.0.0.1]:19080"), "127.0.0.1:19080");
```
`cargo fmt` in that file and push should clear Lint.

### No action needed for now
cargo test: not this PR. The only failure is `chatmail-storage::maildir_cache::tests::store_after_invalidate_is_discarded` (`fresh listing should hit` at `maildir_cache.rs:279`). That test is from #115 (mtime/cache timing). This branch does not touch `chatmail-storage`. The federation / mxdeliv / auth tests in this PR passed. Re-run Test after the fmt push; if that one test fails again it is a flake on `main`, not a federation regression.

Audit and landing already passed.

13sizdahom:
@KamyarAT Thanks — Lint `cargo fmt` fixed and pushed (assert_eq!s collapsed to one line in `crates/chatmail-delivery/src/transport.rs` tests).

Agreed on `cargo test`: the failing `chatmail-storage::maildir_cache::tests::store_after_invalidate_is_discarded` is from #115 / main flake territory; this PR does not touch `chatmail-storage`. Please re-run **Test** after this fmt push.

KamyarAT:
I bumped h2 on this branch (3ffe9cd) so cargo audit would stop failing. That advisory is [RUSTSEC-2026-0258](https://rustsec.org/advisories/RUSTSEC-2026-0258), on h2 0.4.14, published after we rebased onto main. Not from the federation changes; main had the same pin.

You don’t need to do anything else for audit. Lint/test were already green after the rustfmt fix. Once this CI run finishes clean, we’re good to merge.

PR #150, opened 2026-08-19
feat(dkim): enhance DKIM functionality with check and status commands

omidz4t:
Follow-up to `dkim show`: DNS check/status, admin API, and a DKIM tab in admin-web.

Fixes the cargo-audit finding on hickory-proto 0.24.4 (**RUSTSEC-2026-0119**) by bumping `hickory-resolver` to **0.26.1** (rewritten `lookup_txt` for the 0.26 API). Not ignored.

## Summary

- `madmail dkim check` / `GET /admin/dkim/check` compare published `default._domainkey` to the local TXT (does **not** mint a key).
- `madmail dkim status` / `GET /admin/dkim/status` summarize local key + DNS without writing.
- `--json` still prints the data envelope, then **exits 1** when `checked && !matched` (or `lookup_error`).
- First `make run` copies `chatmail.toml.example` if `./data/chatmail.toml` is missing. That file uses **unprivileged** ports (SMTP **2525**, IMAP **1143**, submission **1025**, HTTP **8080**), not production :25.

## Type of change

- [x] Bug fix (audit bump, check-vs-show, JSON exit code)

## Area

- [x] Admin API / admin web UI (`external/madmail-admin-web`)
- [x] Configuration / CLI

## Testing

- [x] `make lint` (fmt, check, clippy `-D warnings`, rustdoc)
- [x] `cargo test -p chatmail-delivery dkim`
- [x] `cargo audit` (exit 0; RUSTSEC-2026-0119 gone)
- [x] E2E: IP skip, missing key does not mint, mismatch `--json` exits 1

## Documentation

- [x] Updated CLI JSON / TDD admin API / admin-web federation DKIM docs

## Security & privacy

- [x] Security-sensitive — cargo audit bump; DKIM TXT `p=` compare; CORS extras stay on `CHATMAIL_ADMIN_CORS_ORIGINS`

## Checklist

- [x] Submodule pointer updated (`external/madmail-admin-web` Check DNS)
- [x] Rebased onto current `main`


crates/chatmail-delivery/Cargo.toml
```diff
@@ -26,6 +26,7 @@ rsa = { version = "0.9", features = ["pem"] }
 base64 = "0.22"
 rand = "0.8"
 viadkim = "0.2"
+hickory-resolver = { version = "0.24", default-features = false, features = ["tokio-runtime", "system-config"] }
```
KamyarAT: Cargo audit is red on this: hickory-proto 0.24.4 is RUSTSEC-2026-0119 (CPU exhaustion in name compression). CI wants >=0.26.1.

Unlike the viadkim/idna pin, there is a patched release, so this should be a bump, not an ignore. 0.26 also dropped TokioAsyncResolver (see lookup_txt), so a version-only change will not compile.

crates/chatmail-delivery/src/dkim.rs
```diff
@@ -110,6 +115,268 @@ pub fn public_txt_path(state_dir: &Path, selector: &str) -> PathBuf {
...
+
+pub fn dkim_txt_matches(expected: &str, found: &[String]) -> bool {
+    let want = normalize_dkim_txt(expected);
+    !want.is_empty() && found.iter().any(|f| normalize_dkim_txt(f) == want)
+}
+
+/// TXT lookup for `name` (FQDN). Empty vec = NXDOMAIN / no TXT.
+pub async fn lookup_txt(name: String) -> Result<Vec<String>, String> {
+    use hickory_resolver::config::{ResolverConfig, ResolverOpts};
+    use hickory_resolver::TokioAsyncResolver;
```
KamyarAT: This is the 0.24 API. After the audit bump to hickory-resolver 0.26, TokioAsyncResolver / tokio_from_system_conf / ResolverConfig::cloudflare() are gone.

0.26 is Resolver::builder_tokio() / TokioResolver, and ResolveErrorKind moved. Worth rewriting this helper in the same commit as the Cargo.toml bump so CI can actually go green.

crates/chatmail/src/ctl/dkim.rs
```diff
@@ -42,85 +38,183 @@ fn show(args: &Args) -> Result<()> {
...
+
+async fn check(args: &Args) -> Result<()> {
+    let ctx = CtlContext::from_args(args)?;
+    let out = CtlOut::from_args(args, "dkim check");
+    let registration = ctx.config.effective_registration_domain(None);
+    let data = check_dns(&ctx.state_dir, &registration)
+        .await
+        .map_err(ChatmailError::config)?;
+
+    if out.is_json() {
```
KamyarAT: `--json` always succeeds here. You emit the payload and return Ok before looking at matched / lookup_error.

Human mode below correctly exits 1 on mismatch or lookup failure. `dkim check --json` is the scripting form, and dkim-check.md / json-output.md both say that case is exit 1.

Please fail the process when checked && !matched (or lookup_error is set). If you return Err after emit, main will also print ok:false on stderr — either emit then exit 1 without the error envelope, or skip emit and return Err. An e2e that asserts the exit code on a mocked mismatch would lock this in.

docs/guide/cli/json-output.md
```diff
@@ -570,6 +570,53 @@ With `-o file`: file contains JSON array (unchanged). Without `-o` and `--json`:
...
+    "dns_fqdn": "default._domainkey.example.org",
+    "expected_txt": "v=DKIM1; k=rsa; p=MIIBIjANBgkqh…",
+    "dns_txt": ["v=DKIM1; k=rsa; p=MIIBIjANBgkqh…"],
+    "matched": true,
+    "checked": true
+  }
+}
+```
+
+`checked` is `false` (no DNS query) for IP-literal domains; then `reason` is set. On mismatch `dns_txt` may be empty. Lookup failures add `lookup_error` and the CLI exits non-zero unless `checked` is false.
```
KamyarAT: This sentence is what `--json` does not do today. ctl/dkim.rs returns Ok right after emit, so lookup_error still comes out as ok:true / exit 0.

Same fix as the check() early return: either make the binary match this paragraph, or change the paragraph (and the exit table in dkim-check.md) if the intended contract is “always 0, read data.matched”.

docs/guide/cli/dkim-check.md
```diff
@@ -0,0 +1,50 @@
...
+
+## Notes
+
+Uses the system resolver (Cloudflare `1.1.1.1` if `/etc/resolv.conf` cannot be read). Concatenates split TXT strings and ignores quotes/whitespace. The RSA `p=` value is compared case-sensitively.
+
+| Result | Exit | Meaning |
+|--------|------|---------|
+| `matched: true` | 0 | DNS TXT matches `{state_dir}/dkim/default.txt` |
+| `checked: false` | 0 | Mail domain is an IP literal — signing is skipped |
+| `matched: false` | 1 | No TXT, wrong key, or lookup error |
```
KamyarAT, line 28: This row is true for human output, false for `--json`. The JSON path never hits the Err returns, so matched:false is still exit 0.

Please keep this table honest once the CLI is fixed (or split human vs --json if you really want JSON to always be 0).

crates/chatmail-delivery/src/dkim.rs
```diff
@@ -110,6 +115,268 @@ pub fn public_txt_path(state_dir: &Path, selector: &str) -> PathBuf {
...
+pub async fn check_dns_with<F, Fut>(
+    state_dir: &Path,
+    primary_domain: &str,
+    lookup: F,
+) -> Result<Value, String>
+where
+    F: FnOnce(String) -> Fut,
+    Fut: std::future::Future<Output = Result<Vec<String>, String>>,
+{
+    let info = publish_info(state_dir, primary_domain)?;
```
KamyarAT: check creates a key. publish_info is the show/install path; status uses inspect_info and does not write.

A first `dkim check` (or GET /admin/dkim/check) on a host with no key will mint default.private, then fail DNS. Please load an existing key only, and return checked:false + a reason when the TXT is missing — same as status.

docs/madmail-admin-web/13-federation-dkim.md
```diff
@@ -0,0 +1,48 @@
...
+Show the outbound federation **DKIM** record so the operator can publish a TXT at `default._domainkey`. Same payload as `madmail dkim show`.
+
+Visiting this page calls `GET /admin/dkim`, which creates `{state_dir}/dkim/default.private` when missing and the mail domain is a DNS name.
+
+## Data loaded
+
+| When | Loader | API |
+|------|--------|-----|
+| Prefetch on this route | `loadDkim()` | `GET /admin/dkim` |
+| Check DNS | `api.dkimCheck()` | `GET /admin/dkim/check` |
```
KamyarAT, line 16: The admin-web commit wires api.dkimCheck / api.dkimStatus, but federation/dkim/+page.svelte never calls them. This table (and the Check DNS button in Actions below) describes UI that is not in the page.

The page is show/copy only. Either add the button, or drop these rows so the madmail docs match the submodule.

docs/madmail-admin-web/13-federation-dkim.md
```diff
@@ -0,0 +1,48 @@
...
+| `dns_fqdn` | Absolute name to publish, or `null` for IP-literal domains |
+| `txt` | TXT value (`v=DKIM1; k=rsa; p=…`), or `null` when not publishable |
+| `publishable` | `true` when a DNS `d=` exists and the key is on disk |
+| `generated` | `true` when this request created the key |
+| `reason` | Why the record is not publishable (IP-literal domains) |
+| `private_key_path` / `txt_path` | Paths on the server |
+
+## Actions
+
+Copy buttons for FQDN, TXT, and a BIND-style zone line. **Check DNS** calls `GET /admin/dkim/check` (`madmail dkim check`) and shows whether the published TXT matches.
```
KamyarAT, line 35: There is no Check DNS control in +page.svelte — only copy buttons for FQDN / TXT / zone line.

Also: store.refresh() in the submodule always calls loadDkim(), so a generic refresh can create default.private without opening this tab. Header refresh is already scoped to /federation/dkim; global refresh should be too.

crates/chatmail-admin/src/cors.rs
```diff
@@ -42,9 +42,27 @@ fn configured_origins() -> Vec<String> {
 }
 
 pub fn is_allowed_origin(origin: &str) -> bool {
+    if is_loopback_dev_origin(origin) {
```
KamyarAT: This is a production CORS policy change in a DKIM PR. Any http(s) localhost origin is now allowed, any port.

The admin-web Vite config already proxies /api/admin to loopback, so the browser stays same-origin for the intended dev path. Extra origins belong on CHATMAIL_ADMIN_CORS_ORIGINS, not hardcoded for every madmail install.

crates/chatmail-admin/src/cors.rs
```diff
@@ -42,9 +42,27 @@ fn configured_origins() -> Vec<String> {
...
+/// Vite / Electron on this machine (`http://localhost:5173`, `http://127.0.0.1:…`).
+/// LAN IPs are not included — use the Vite proxy so madmail stays loopback-only.
+fn is_loopback_dev_origin(origin: &str) -> bool {
+    let Some(rest) = origin
+        .strip_prefix("http://")
+        .or_else(|| origin.strip_prefix("https://"))
+    else {
+        return false;
+    };
+    let Some(host) = rest.split('/').next().and_then(|h| h.split(':').next()) else {
```
KamyarAT: This does not parse IPv6. Origin http://[::1]:5173 becomes rest="[::1]:5173", split(':').next() is "[", and the "[::1]" arm never matches.

The tests only cover localhost / 127.0.0.1. If loopback origins stay, parse host/port properly (or drop [::1] from the list).

crates/chatmail-delivery/src/dkim.rs
```diff
@@ -110,6 +115,268 @@ pub fn public_txt_path(state_dir: &Path, selector: &str) -> PathBuf {
...
+                    k.to_ascii_lowercase(),
+                    v.to_ascii_lowercase()
+                ))
+            }
+        })
+        .collect::<Vec<_>>()
+        .join(";")
+}
+
+pub fn dkim_txt_matches(expected: &str, found: &[String]) -> bool {
```
KamyarAT, line 300: This is an exact compare after normalize, so tag order and extra tags fail the check.

v=DKIM1; p=…; k=rsa or a DNS UI adding t=y will miss even when p= is the local key. Fine if operators paste show output unchanged; brittle otherwise. Comparing a parsed tag map (p= case-sensitive, other tags case-insensitive) would match what filtermail actually cares about.

docs/TDD/09-admin-api.md
```diff
@@ -79,6 +82,41 @@ Madmail-compatible JSON-RPC admin API. Full operator reference: [`context/madmai
...
+
+Other methods: HTTP 405.
+
+### `/admin/dkim/status`
+
+`GET` summarizes local key presence and DNS match (`madmail dkim status`). Does not create a key. Body includes `key_present`, `publishable`, `dns_checked`, `dns_matched`.
+
+Other methods: HTTP 405.
+
+Toggle POST body: `{"action": "enable"}` or `{"action": "disable"}`.
```
KamyarAT: Stray copy: the Toggle POST body line is already above the new DKIM sections (line 83). This one landed in the middle of Push/Setting and is not about DKIM. Please delete it.

Makefile
```diff
@@ -320,11 +320,19 @@ docker-down:
...
+ensure-dev-config:
+	@mkdir -p $(STATE_DIR)
+	@if [ ! -f $(CONFIG) ]; then \
+		cp chatmail.toml.example $(CONFIG); \
+		echo "Wrote $(CONFIG) (unprivileged ports: SMTP 2525, IMAP 1143, submission 1025, HTTP 8080)"; \
+	fi
+
 # ── Run local server ─────────────────────────────────────────────────────────
-run: build
+run: build ensure-dev-config
```
KamyarAT, line 332: Fine for local DX, but first `make run` now seeds unprivileged 2525/1143/1025/8080 instead of production :25.

Worth a sentence in the PR body so nobody thinks the server is bound on the usual ports. The copy-if-missing guard is the right shape.

KamyarAT (changes requested):
Follow-up to `dkim show`: check/status, admin API, and a DKIM tab in admin-web. Direction is right. Not mergeable yet.

Cargo audit is red on hickory-proto 0.24.4 (RUSTSEC-2026-0119). Bump hickory-resolver to >=0.26.1 — do not ignore this one, there is a patched release. 0.26 renamed TokioAsyncResolver, so lookup_txt will need a small rewrite.

`madmail dkim check --json` always exits 0. Human mode correctly fails on mismatch/lookup error; JSON returns early with ok:true. That disagrees with dkim-check.md and json-output.md. Scripts that only look at the exit code will green a missing TXT. Either fail the process after emitting data.matched, or document that --json is always 0 and add an e2e.

`dkim check` currently calls publish_info, so it will create a key. status is the read-only path. check should load an existing key and skip if there isn't one.

Admin-web: the Federation → DKIM page is show/copy only. api.dkimCheck / dkimStatus are unused, and 13-federation-dkim.md describes a Check DNS button that is not in the page. store.refresh() also loadDkim(), so a generic refresh can mint default.private. Either add the button or cut the docs.

The localhost CORS allow is a separate policy change (Vite already proxies /api/admin). split(':') also breaks [::1]. I'd keep extra origins on CHATMAIL_ADMIN_CORS_ORIGINS.

Please rebase onto current main (Cargo.lock already diverged) and fill in the PR template. chatmail.toml.example for unprivileged make run is welcome — call out the first-run port change in the description.

KamyarAT:
@omidz4t Thanks, the review comments are addressed.

Hickory is 0.26.1 (audit green), check is read-only, --json prints the envelope and exits 1 on mismatch with an empty stderr, TXT compare allows extra/reordered tags, loopback CORS is gone, and admin-web has Check DNS without minting a key from global refresh.

CI is green. Fine to merge.

Tiny leftover, not blocking: dkim-check.md still says Cloudflare 1.1.1.1 for the resolver fallback (code uses Google), and the checked:false exit table only mentions IP literals (missing key is the same skip now).

KamyarAT (approved):

PR #160, opened 2026-09-02
feat: madmail db sqlite-to-postgres

KamyarAT:
## Summary

Adds `madmail db sqlite-to-postgres` to copy application SQL tables from SQLite into PostgreSQL.

This is the copy tool the operator Postgres guide described. It does **not** move Maildir, `sharing.db`, the retry queue, or `admin_token`. After the copy, set `driver postgres` and restart.

Also CAST Postgres `TIMESTAMP` columns to text in token list/status and DNS override reads so the copied DB works with the existing CLI (SQLite stores those as TEXT).

Stacked on #159 (empty-database Postgres how-to).

## Type of change

- [ ] Bug fix (non-breaking change that fixes an issue)
- [ ] Breaking change (fix or feature that would cause existing behavior to change)
- [ ] Documentation only
- [ ] Build / CI / tooling
- [ ] Refactor (no functional change)

## Area

- [ ] SMTP
- [ ] IMAP
- [ ] Federation / delivery queue
- [ ] Authentication / JIT
- [x] Admin API / admin web UI (`external/madmail-admin-web`)
- [x] Storage / quota / DB migrations
- [x] Configuration / CLI
- [ ] Other:

## Testing

- [ ] `cargo fmt --all -- --check`
- [x] `cargo clippy -p chatmail-db -p chatmail -- -D warnings`
- [x] targeted tests: `chatmail-db sqlite_to_postgres`, `chatmail dispatch_db`, `ctl_cli_e2e e2e_ctl_db_sqlite`, endpoint_cache, registration_token

**Manual verification (if any):**

```text
Local madmail binary (not Docker). Postgres in Docker on 127.0.0.1:55432.

S1 dry-run: no Postgres tables created; passwords=2
S2 copy: settings/passwords/blocklist/tokens/dns/modseq/push/exchangers/message_stats
S3 second copy without --force refused
S4 --force after adding carol: passwords=3
S5 second local instance with postgres DSN:
    accounts status login_count=3
    accounts info alice/bob/carol
    blocklist, tokens, endpoint-cache, federation dismiss-list
S6 missing sqlite file errors
S7 postgres-config instance requires --sqlite
S8 Go key/value passwords table → username/hash on Postgres
S9 Maildir stayed on sqlite instance; not copied
S10 JSON command string db sqlite-to-postgres
```

## Documentation

- [ ] No doc updates needed
- [x] Updated operator/user docs (`docs/project/user-guide/`, install guides)
- [x] Updated developer docs (`docs/project/`)
- [x] Updated or added TDD section (`docs/TDD/`)

## Security & privacy

- [x] Not security-sensitive
- [ ] Security-sensitive — added/updated tests for the security property
- [ ] Reviewed error messages for information leakage

## Commits

- `feat:` new feature
- `fix:` timestamp decode on Postgres

## Checklist

- [ ] Submodule pointers updated if `external/madmail-admin-web` changed


mamaru-a:
Issues: 


### 1. High — legacy Go stats columns break the copy

`crates/chatmail-db/src/sqlite_to_postgres.rs:486`

`copy_federation_stats` hardcodes `failed_https` / `success_https`. Legacy Madmail (Go) databases name those columns `failed_http_s` / `success_http_s` — that is exactly what `schema::federation_stats_columns()` exists to resolve, and this copier does not use it.

**Scenario:** an operator migrates a Go-era SQLite DB — the same class of DB whose key/value `passwords` layout `copy_passwords` explicitly supports. `federation_server_stats` exists, so it is not skipped; the SELECT fails with `no such column: failed_https` and the copy aborts at table 9 of 13, after the earlier tables were already written.

### 2. High — legacy singular `quota` table silently not copied

`crates/chatmail-db/src/sqlite_to_postgres.rs:36`

`COPY_TABLES` lists only `quotas`. Legacy databases keep account data in the singular `quota` table — see `schema::quota_table()`, used by `account_info`, `maintenance`, `quota_defaults` and `passwords`.

**Scenario:** a Go-era SQLite DB opened by v2 gets an empty `quotas` created by `apply_legacy_schema_tables`, while the live rows stay in `quota`. The copier reads `quotas` and reports `quotas 0 0 no` — indistinguishable from success, not even marked skipped. Every account's `created_at` / `first_login_at` / `last_login_at` / `max_storage` / `used_token` is lost on Postgres, while passwords copy fine, so the loss is not obvious.

### 3. High — `--force` deletes all 13 tables before reading any source row

`crates/chatmail-db/src/sqlite_to_postgres.rs:129`

The DELETE loop wipes every one of the 13 `COPY_TABLES` in Postgres before a single source row is read, outside any transaction, and tables missing from the source are never repopulated.

Two concrete destroy-then-abort paths:

- **(a) Legacy source.** The DELETE loop empties all 13 tables, then the copy dies on `failed_https` (finding 1), leaving the destination emptied and only half repopulated, with no rollback.
- **(b) Destination-layout blindness.** `run_migrations()` deliberately skips the v2 schema when `legacy_madmail_schema_present(dst)`, yet `copy_passwords` picks its INSERT column shape from the *source* layout and never inspects the destination. Against a Go Postgres DB with key/value `passwords`, `DELETE FROM passwords` succeeds and `INSERT INTO passwords (username, hash, created_at)` then fails — destroying every credential.

Separately, a source missing e.g. `exchangers` leaves that Postgres table wiped and empty.

Suggested fix: read all source tables first, wrap DELETE + INSERT in a single `dst.begin()` transaction, and refuse when `passwords_layout(dst)` does not match the source. The confirm prompt in `ctl/db.rs` also only mentions "refuse if passwords already exist" and never says `--force` deletes every row in 13 tables.

### 4. Medium — `exchangers.last_poll_at` read without CAST

`crates/chatmail-admin/src/resources/exchangers.rs:41`

The exchangers admin GET selects `last_poll_at` (Postgres `TIMESTAMP`) into `Option<String>` with no CAST — the same decode failure this PR fixes for `registration_tokens` and `dns_overrides`. Exchanger rows are now copied to Postgres by this tool, which makes it reachable.

**Scenario:** after `db sqlite-to-postgres` and switching `driver` to postgres, the admin exchangers resource returns 500 for any row with a non-null `last_poll_at` (`mismatched types`). Needs `CAST(last_poll_at AS TEXT)` like the other reads in this PR.

### 5. Medium — non-empty guard checks only the `passwords` table

`crates/chatmail-db/src/sqlite_to_postgres.rs:119`

**Scenario:** the target Postgres DB is already configured (settings rows, registration tokens, `dns_overrides`) but has zero accounts. `pw_count == 0`, so there is no refusal, and every `copy_*` upsert is `ON CONFLICT DO UPDATE` — the existing Postgres settings and tokens are silently replaced by the SQLite ones, with no warning and without `--force`.

### 6. Low — `expires_at` bound as text into a Postgres `TIMESTAMP` (pre-existing, outside this diff)

`crates/chatmail/src/ctl/registration_tokens.rs:90`, also `crates/chatmail-admin/src/resources/tokens.rs:205` and `:229`

`expires_at` is bound as a `String` into a `TIMESTAMP` column, so token writes fail on the Postgres database this PR's copy produces: `column expires_at is of type timestamp without time zone but expression is of type text`. This PR fixes the read side of this class, and its own copier casts with `$5::timestamp` — the write side still needs a `::timestamp` cast or a typed binding.

---

**Checked and cleared:** `CAST(...)` column naming is safe (all row types are tuples, so decoding is positional, not by name); Postgres fractional-second timestamps are fine (`parse_sqlite_timestamp` accepts them via the RFC 3339 branch).

**Deliberately not reported:** dry-run opens the SQLite file twice; the `has_action` branch in `copy_federation_rules` is effectively dead; the `_ => Ok(0)` arm in `copy_one`.

mamaru-a (commented):
**Medium — `crates/chatmail-admin/src/resources/exchangers.rs:41`.** The exchangers admin GET selects `last_poll_at` (Postgres `TIMESTAMP`) into `Option<String>` with no CAST — the same decode failure this PR fixes for `registration_tokens` and `dns_overrides`. Exchanger rows are now copied to Postgres by this tool, which is what makes it reachable: after switching `driver` to postgres, the endpoint returns 500 for any row with a non-null `last_poll_at` (`mismatched types`). Needs `CAST(last_poll_at AS TEXT)` like the other reads here.

**Low, pre-existing — `crates/chatmail/src/ctl/registration_tokens.rs:90`, plus `crates/chatmail-admin/src/resources/tokens.rs:205` and `:229`.** `expires_at` is bound as a `String` into a `TIMESTAMP` column, so token writes fail on the Postgres database this copy produces: `column expires_at is of type timestamp without time zone but expression is of type text`. This PR fixes the read side of that class, and its own copier casts with `$5::timestamp` — the write side still needs a `::timestamp` cast or a typed binding.

---

**Checked and cleared:** `CAST(...)` column naming is safe (all row types are tuples, so decoding is positional, not by name); Postgres fractional-second timestamps are fine (`parse_sqlite_timestamp` accepts them via the RFC 3339 branch).

**Deliberately not reported:** dry-run opens the SQLite file twice; the `has_action` branch in `copy_federation_rules` is effectively dead; the `_ => Ok(0)` arm in `copy_one`.

crates/chatmail-db/src/sqlite_to_postgres.rs
```diff
@@ -0,0 +1,691 @@
...
+        i64,
+        i64,
+        i64,
+        i64,
+        i64,
+        i64,
+    )> = crate::db_fetch_all!(
+        src,
+        (String, i64, i64, i64, i64, i64, i64, i64, i64, i64, i64, i64),
+        "SELECT domain, queued_messages, failed_http, failed_https, failed_smtp, \
```
mamaru-a: **High — legacy Go stats columns break the copy.**

This hardcodes `failed_https` / `success_https`, but legacy Madmail (Go) databases name those columns `failed_http_s` / `success_http_s`. That is exactly what `schema::federation_stats_columns()` exists to resolve, and it is not used here.

**Scenario:** an operator migrates a Go-era SQLite DB — the same class of DB whose key/value `passwords` layout `copy_passwords` explicitly supports. `federation_server_stats` exists, so it is not skipped; the SELECT fails with `no such column: failed_https` and the copy aborts at table 9 of 13, *after* the earlier tables were already written.

Suggest resolving the column names via `federation_stats_columns(src)` and building the SELECT from them.

crates/chatmail-db/src/sqlite_to_postgres.rs
```diff
@@ -0,0 +1,691 @@
...
+use sqlx::postgres::PgPool;
+use sqlx::Row;
+
+use crate::pool::{connect_database, run_migrations, DbPool};
+use crate::schema::{passwords_layout, table_exists, PasswordsLayout};
+
+/// Tables copied in this order (no cross-table FKs).
+pub const COPY_TABLES: &[&str] = &[
+    "settings",
+    "quotas",
```
mamaru-a, line 38: **High — legacy singular `quota` table silently not copied.**

`COPY_TABLES` lists only `quotas`. Legacy databases keep account data in the singular `quota` table — see `schema::quota_table()`, used by `account_info`, `maintenance`, `quota_defaults` and `passwords`.

**Scenario:** a Go-era SQLite DB opened by v2 gets an empty `quotas` created by `apply_legacy_schema_tables`, while the live rows stay in `quota`. The copier reads `quotas` and reports `quotas 0 0 no` — indistinguishable from success, and not even marked skipped. Every account's `created_at` / `first_login_at` / `last_login_at` / `max_storage` / `used_token` is lost on Postgres, while passwords copy fine, so the loss is not obvious.

Suggest resolving the source table via `quota_table(src)` before reading.

crates/chatmail-db/src/sqlite_to_postgres.rs
```diff
@@ -0,0 +1,691 @@
...
+    if pw_count > 0 && !opts.force {
+        return Err(ChatmailError::config(format!(
+            "Postgres passwords table already has {pw_count} row(s). \
+             Refusing to overwrite (pass --force to replace)."
+        )));
+    }
+
+    if opts.force {
+        for name in COPY_TABLES.iter().rev() {
+            sqlx::query(&format!("DELETE FROM {name}"))
```
mamaru-a, line 174: **High — `--force` deletes all 13 tables before reading any source row.**

The DELETE loop wipes every one of the 13 `COPY_TABLES` in Postgres before a single source row is read, outside any transaction, and tables missing from the source are never repopulated.

Two concrete destroy-then-abort paths:

- **(a) Legacy source.** The loop empties all 13 tables, then the copy dies on `failed_https` (see the comment on `copy_federation_stats`), leaving the destination emptied and only half repopulated, with no rollback.
- **(b) Destination-layout blindness.** `run_migrations()` deliberately skips the v2 schema when `legacy_madmail_schema_present(dst)`, yet `copy_passwords` picks its INSERT column shape from the *source* layout and never inspects the destination. Against a Go Postgres DB with key/value `passwords`, `DELETE FROM passwords` succeeds and `INSERT INTO passwords (username, hash, created_at)` then fails — destroying every credential.

Separately, a source missing e.g. `exchangers` leaves that Postgres table wiped and empty.

Suggested fix: read all source tables first, wrap DELETE + INSERT in a single `dst.begin()` transaction, and refuse up front when `passwords_layout(dst)` does not match the source. The confirm prompt in `ctl/db.rs` should also say that `--force` deletes every row in 13 tables — right now it only mentions "refuse if passwords already exist".

crates/chatmail-db/src/sqlite_to_postgres.rs
```diff
@@ -0,0 +1,691 @@
...
+
+    let dst_pool = connect_database(&DatabaseConfig {
+        driver: DbDriver::Postgres,
+        dsn: postgres_dsn.to_string(),
+    })
+    .await?;
+    run_migrations(&dst_pool).await?;
+    let dst = pg_pool(&dst_pool)?;
+
+    let pw_count = count_pg(dst, "passwords").await?;
```
mamaru-a: **Medium — the non-empty guard checks only the `passwords` table.**

**Scenario:** the target Postgres DB is already configured (settings rows, registration tokens, `dns_overrides`) but has zero accounts. `pw_count == 0`, so there is no refusal, and every `copy_*` upsert is `ON CONFLICT DO UPDATE` — the existing Postgres settings and tokens are silently replaced by the SQLite ones, with no warning and without `--force`.

Suggest counting across all `COPY_TABLES` (or at least `settings` / `quotas` / `registration_tokens`) and naming the non-empty ones in the refusal message.

mamaru-a:
Pushed `221bac1` addressing all six findings from the review above.

| Finding | Fix |
|---|---|
| Legacy `failed_http_s` / `success_http_s` | `copy_federation_stats` now resolves names via `schema::federation_stats_columns` |
| Legacy singular `quota` | new `source_table()` resolves via `schema::quota_table`; `TableCopy` reports `source_table` so a legacy read is visible |
| `--force` deletes before reading, no transaction | whole copy (deletes included) runs inside one `pg.begin()`; destination unchanged on any error |
| — same, destination-layout blindness | refuses up front when the destination `passwords` still uses the Madmail key/value layout |
| Guard only checked `passwords` | now counts every copied table and names the non-empty ones; CLI confirmation states what `--force` deletes |
| `last_poll_at` / `expires_at` timestamps | `CAST(last_poll_at AS TEXT)` on the admin read; `expires_at` bound through a new `DbPool::timestamp_param()` on all three write paths |

Added `inspect_prefers_legacy_quota_table` as a regression test for the `quota` case.

Verified: `cargo fmt --all -- --check`, `cargo clippy -p chatmail-db -p chatmail-admin -p chatmail --all-targets -- -D warnings`, and `chatmail-db sqlite_to_postgres` / `registration_token`, `chatmail-admin tokens`, `chatmail dispatch_db`, `ctl_cli_e2e e2e_ctl_db` — all green. Run on Linux in Docker, since `chatmail-admin` does not build on macOS (pre-existing `u32`/`u64` errors in `status_storage.rs`, present without these changes).

The Postgres-specific paths (the transaction rollback, the layout refusal, the timestamp casts) are not covered by a test — they need a live Postgres, which the suite does not have.

mamaru-a:
Also pushed `4d3039a`, fixing the macOS build break I hit while verifying the review fixes.

`statvfs` field widths are target-dependent: on macOS `f_blocks` / `f_bavail` are `u32` while `f_frsize` is `u64`, so `stat.f_blocks * bsize` in `resources/status_storage.rs` failed with `cannot multiply u32 by u64`. On Linux all three are `u64`, which is why CI never caught it — but it broke `cargo check` for `chatmail-admin` and every crate depending on it on any macOS checkout. All three fields are now widened to `u64`, with `saturating_mul` instead of `*`.

Pre-existing and unrelated to this PR's feature — happy to split it into its own PR off `main` if you would rather keep this branch focused.

`cargo check --workspace --all-targets` now passes on macOS (it did not before), and `cargo fmt --check` plus `cargo clippy -p chatmail-admin --all-targets -- -D warnings` and the `status_storage` tests are green on Linux.

mamaru-a:
now it's fine to be merged. @KamyarAT

ArianAr:
@mamaru-a Thanks for the review and the follow-up commits.

macOS disk_usage (`4d3039a`): keep it on this branch. It’s a real compile break , Linux CI never sees it, and splitting it off main isn’t worth the extra round-trip.

Live Postgres: agreed the suite is SQLite-only, so the dest-side paths (transaction / `--force` order, KV `passwords` refusal, timestamp binds) never run in CI. We’ll add a small Postgres service on the Test job and skip those tests when `TEST_POSTGRES_DSN` is unset. Scope is just those dest behaviours: refuse without `--force` leaves existing rows, refuse Go key/value `passwords` before any DELETE, and a happy-path copy we can SELECT back (including `expires_at` / `last_poll_at`). Not a full operator matrix. That’ll land on this PR before merge.

KamyarAT:
@mamaru-a Thanks for the review and the follow-up commits.

macOS disk_usage (`4d3039a`): keep it on this branch. It’s a real compile break , Linux CI never sees it, and splitting it off main isn’t worth the extra round-trip.

Live Postgres: agreed the suite is SQLite-only, so the dest-side paths (transaction / `--force` order, KV `passwords` refusal, timestamp binds) never run in CI. We’ll add a small Postgres service on the Test job and skip those tests when `TEST_POSTGRES_DSN` is unset. Scope is just those dest behaviours: refuse without `--force` leaves existing rows, refuse Go key/value `passwords` before any DELETE, and a happy-path copy we can SELECT back (including `expires_at` / `last_poll_at`). Not a full operator matrix. That’ll land on this PR before merge.

KamyarAT:
Pushed dest-side coverage for the copy path.

The Test job now starts `postgres:16` and sets `TEST_POSTGRES_DSN`. Those tests skip when the env is unset, so a checkout without Postgres still runs the rest of the suite.

What they actually hit:

• refuse without `--force` leaves dest rows in place
• Go key/value `passwords` is refused before any `DELETE`, even with `--force`
• a full copy of the application tables, including `expires_at` / `last_poll_at` and a dest-side timestamp write
• `--force` really replaces dest-only rows
• legacy SQLite `quota` lands in `quotas`
• Go `failed_http_s` / `success_http_s` land in the v2 column names

One behaviour that only showed up against a real dest: sqlx seeds three zero `message_stats` counters, so a brand-new database always looked occupied. The non-empty guard now skips that table; the copy still upserts the counters. Without `--force` you still cannot overwrite settings, passwords, tokens, etc.

mamaru-a:
perfection. 
looks good, ill merge

Issue #19, opened 2026-01-21
[TASK] Real-time Server Monitoring (Online users & Throughput)


omidz4t:
# 📋 Overview
Implement real-time monitoring capabilities for the Madmail server to track live performance metrics. Initially, these metrics should be accessible via the CLI, with the intention of later exposing them for graphical dashboarding in the web UI.

## 👤 User Story / Motivation
- **As a:** Madmail Server Administrator
- **I want to:** See the current number of online users and the message throughput (messages per second).
- **So that:** I can monitor the server's health, understand resource requirements, and identify potential spikes or issues as they happen.

---

## ✅ Functional Requirements
- [ ] **Online User Tracking:** Implement a mechanism to count active/concurrent IMAP and SMTP sessions.
- [ ] **Throughput Monitoring:** Track the number of messages successfully processed per second/minute.
- [ ] **CLI Monitoring Command:** Create a new command `maddy monitor` (or similar) that displays these real-time stats.
- [ ] **Data Export Readiness:** Ensure the internal metrics are exposed in a way that can be easily converted to a web-facing API or chart data later.

---

## 🛠 Proposed Implementation
- **Backend Tracking:** Add global counters or session registries in the SMTP and IMAP endpoint packages.
- **CLI Implementation:** Add a new subcommand in `internal/cli/ctl/` that queries the running server (potentially via a control socket or internal API) to fetch these live stats.
- **Reporting:** Use a moving average or a sliding window to calculate "messages per second".

---

## 🧪 Verification & Testing
- [ ] **Live Monitoring:** Start multiple sessions and verify the CLI correctly reports the active user count.
- [ ] **Performance Benchmarking:** Use the stress test suite (#9) and verify that the throughput metrics match the actual load being generated.

---

## 📅 Roadmap/Checklist
- [ ] Metric tracking implemented in core.
- [ ] `maddy monitor` command added.
- [ ] Conventional Commits used.
- [ ] Definition of Done reached.

mamaru-a:
## Implementation Plan

### Approach: reuse the Prometheus registry we already have

The repo already ships a full metrics stack — `github.com/prometheus/client_golang` with counters in `internal/endpoint/smtp/metrics.go`, `internal/msgpipeline/metrics.go`, `internal/target/queue/metrics.go`, and an HTTP exporter endpoint in `internal/endpoint/openmetrics/om.go` serving `/metrics`.

[...]

PR #163, opened 2026-09-07
feat(metrics): track concurrent IMAP and SMTP connections

mamaru-a:
Step 1 of #19: the "online users" metric.

## What

New `internal/connmetrics` package: a `net.Listener` wrapper that reports currently open client connections as a Prometheus gauge, `maddy_conns_active{module}`, labelled with the endpoint instance name. Wired into the IMAP and SMTP endpoints, plus the chatmail ALPN path.

It is exported through the existing `openmetrics` endpoint, so nothing new is needed to read it. Seven lines of wiring; the rest is the package and its test.

## Placement, and why it is not where the issue plan said

The plan comment on #19 proposed wrapping inside `Endpoint.Serve(l net.Listener)`. That turned out to be wrong twice:

- `setupListeners` calls `endp.serv.Serve(l)` directly, never `endp.Serve(l)`, so the production listeners would not have been counted at all.
- It puts the wrapper *above* TLS. go-imap reads the TLS state of an implicit-TLS connection with a bare `tlsConn, _ := c.(*tls.Conn)` (`server/conn.go:88`). A wrapper there makes that assertion fail, `TLSState()` returns nil, and an IMAPS connection looks like plaintext to the auth checks — silently, with no error anywhere.

So the wrap goes innermost, right after `net.Listen` and below `tls.NewListener`.

`internal/endpoint/chatmail` is a third surface the plan missed: it serves IMAP and SMTP over an ALPN-multiplexed TLS port and hands the endpoints conns that are already `*tls.Conn`. Same rule, so it accounts for those with `connmetrics.WrapConn` on the raw buffered conn *before* `tls.Server`.

`Close` is guarded with a `sync.Once` — both go-imap and go-smtp close a connection more than once on some paths, and an unguarded `Dec` would drift the gauge negative.

## Scope

This counts concurrent connections, which is the working definition of "session" for IMAP and SMTP and what an admin needs for capacity planning. It does not track distinct authenticated identities; if we want that later, it is an additive `map[string]int` hooked into the login paths, no rework here.

Throughput (step 2), the `openmetrics` config default (step 3) and `maddy monitor` (step 4) are separate PRs.

## Verification

- `go build ./internal/...` and `go vet ./internal/...` clean.
- New test in `internal/connmetrics`: accepts N connections, asserts the gauge reads N, closes each one twice, asserts it returns to 0.
- `go test ./internal/endpoint/...` fails on the same 12 tests before and after this change (pre-existing, `PGP verification failed: mime: no media type`), confirmed against a stashed baseline.
- Grepped go-smtp, go-imap and `internal/` for `*net.TCPConn` / `SetKeepAlive` / `SetNoDelay` — no hits, so nothing else loses a capability by sitting above the wrapper.
- `go.mod` gains one indirect dependency (`kylelemons/godebug`, pulled in by `prometheus/testutil`); `go.sum` is unchanged.

Note for step 4: a series does not exist until that endpoint accepts its first connection, so the CLI must render a missing series as 0 rather than erroring.

I could not run `golangci-lint` locally — the installed binary is v2 and `.golangci.yml` is v1 format (`can't load config: unsupported version of the configuration`). errcheck was satisfied by hand instead.

### DIFF
--- a/go.mod
+++ b/go.mod
@@ -203,6 +203,7 @@ require (
 	github.com/klauspost/cpuid/v2 v2.2.9 // indirect
 	github.com/kulti/thelper v0.6.3 // indirect
 	github.com/kunwardeep/paralleltest v1.0.10 // indirect
+	github.com/kylelemons/godebug v1.1.0 // indirect
 	github.com/lasiar/canonicalheader v1.1.2 // indirect
 	github.com/ldez/exptostd v0.4.2 // indirect
 	github.com/ldez/gomoddirectives v0.6.1 // indirect
--- a/internal/connmetrics/listener.go
+++ b/internal/connmetrics/listener.go
@@ -1,0 +1,87 @@
+/*
+Maddy Mail Server - Composable all-in-one email server.
+Copyright © 2019-2020 Max Mazurov <fox.cpp@disroot.org>, Maddy Mail Server contributors
+
+This program is free software: you can redistribute it and/or modify
+it under the terms of the GNU General Public License as published by
+the Free Software Foundation, either version 3 of the License, or
+(at your option) any later version.
+
+This program is distributed in the hope that it will be useful,
+but WITHOUT ANY WARRANTY; without even the implied warranty of
+MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.  See the
+GNU General Public License for more details.
+
+You should have received a copy of the GNU General Public License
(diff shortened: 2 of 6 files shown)


mamaru-a:
Closing: this was written against the pre-Rust Go tree, which no longer exists on main. Redoing the connection gauge against `crates/chatmail-metrics` and the `chatmail-imap`/`chatmail-smtp` accept loops instead. Refs #19

PR #164, opened 2026-09-07
feat(metrics): track concurrent IMAP and SMTP connections

mamaru-a:
> **Stack:** themadorg/madmail#164 (step 1, the gauge) → themadorg/madmail#165 (step 4, `madmail monitor`). #165 is based on the `feat/19-conn-metrics` branch, so once this PR merges it needs retargeting to `main`. Rebased onto 2.27.0; no code changes in the rebase.

Step 1 of #19: the "online users" metric. Replaces #163, which was written against the pre-Rust tree.

## What

`maddy_conns_active{module}` — an `IntGaugeVec` in `chatmail-metrics`, held up for the lifetime of every accepted IMAP and SMTP connection. Exported through the existing `openmetrics` listener, so nothing new is needed to read it.

Three lines of wiring; the rest is the guard and its tests.

## Why an RAII guard and not open/close calls

The count is released on `Drop`, not by an explicit close call. A connection task can end in more ways than the happy path — a failed TLS handshake returns early, a session can panic, and the runtime drops in-flight tasks at shutdown. A gauge that misses any of those only ever climbs, which makes it worse than useless for the monitoring this exists for.

`conn_guard` is `#[must_use]`: `let _ = conn_guard(..)` drops the guard on the same line and would silently count nothing. The binding has to be `let _guard = ...`.

The guard is created as the first statement inside `tokio::spawn`, so the TLS-handshake-failure branch is covered too.

## Notes

- Label children for `imap`, `smtp` and `submission` are pre-created in `init_metrics()`, matching what the SMTP counters already do there, so the series is present and reads `0` before the first connection instead of being absent. That matters for the CLI in step 4 — it should still handle a missing series as 0, but it will not normally hit that.
- `chatmail-imap`'s existing `connection_stats` module is untouched. It also tracks per-peer IPs for a different consumer and is not interchangeable with a gauge; duplicating the count is cheaper than entangling the two.
- SMTP labels itself from `cfg.module`, so `smtp` and `submission` separate on their own, consistent with the existing SMTP counters.
- `chatmail-imap` gains a dependency on `chatmail-metrics` (a leaf crate — no cycle). `chatmail-smtp` already had it.

## Scope

Concurrent connections, which is the working definition of "session" for IMAP and SMTP and what an admin needs for capacity planning. Not distinct authenticated identities — if we want that later it is additive, and `connection_stats` already has the per-IP half of it.

Throughput and the `monitor` CLI are separate; see the plan comments on #19, which I will re-scope against the Rust crates now that this one is done.

## Verification

- `cargo build -p chatmail-metrics -p chatmail-smtp -p chatmail-imap` — clean.
- `cargo clippy --all-targets -- -D warnings` on the three crates — clean. `cargo fmt --all --check` — clean. `RUSTDOCFLAGS="-D warnings" cargo doc --no-deps` — clean.
- `cargo test -p chatmail-metrics` — 7 passed, including two new ones: N guards read N and drop to 0, and a guard inside `catch_unwind` that panics still leaves the gauge at 0. The panic case is the one that proves the RAII choice earned its lines. Both use unique label values, since the statics are global and tests share a process.
- `cargo test -p chatmail-imap -p chatmail-smtp`: `p6_imap_idle_unsolicited_exists` fails, but it fails identically on a clean checkout of this branch point (verified by stashing) — pre-existing, unrelated.

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -877,6 +877,7 @@ dependencies = [
  "chatmail-config",
  "chatmail-db",
  "chatmail-iroh",
+ "chatmail-metrics",
  "chatmail-pgp",
  "chatmail-push",
  "chatmail-state",
--- a/crates/chatmail-imap/Cargo.toml
+++ b/crates/chatmail-imap/Cargo.toml
@@ -10,6 +10,7 @@ chatmail-config = { workspace = true }
 chatmail-db = { workspace = true }
 chatmail-turn = { workspace = true }
 chatmail-iroh = { workspace = true }
+chatmail-metrics = { workspace = true }
 chatmail-push = { workspace = true }
 chatmail-pgp = { workspace = true }
 chatmail-state = { workspace = true }
--- a/crates/chatmail-imap/src/server.rs
+++ b/crates/chatmail-imap/src/server.rs
@@ -66,6 +66,7 @@ pub async fn run_imap_listener(
                 let cfg = cfg.clone();
                 let acceptor = tls_acceptor.clone();
                 tokio::spawn(async move {
+                    let _conn_guard = chatmail_metrics::conn_guard("imap");
                     connection_stats::on_open(&peer_ip);
                     let mut session = ImapSession::new(ctx, pool, cfg);
                     let result = if let Some(acceptor) = acceptor {
--- a/crates/chatmail-metrics/src/lib.rs
+++ b/crates/chatmail-metrics/src/lib.rs
@@ -21,7 +21,8 @@ mod metrics;
 mod server;
 
 pub use metrics::{
-    exposition_text, init_metrics, record_smtp_aborted, record_smtp_completed,
+    conn_guard, exposition_text, init_metrics, record_smtp_aborted, record_smtp_completed,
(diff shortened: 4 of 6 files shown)


KamyarAT (approved):
Reviewed this as the first PR in the stack (not #165).

The gauge is the right shape: RAII `conn_guard` as the first statement in the IMAP/SMTP `tokio::spawn` accept tasks, `cfg.module` for smtp vs submission, labels pre-created in `init_metrics`, imap → metrics is a leaf dep. `let _conn_guard` is the binding that actually keeps the count; TLS handshake failure still drops the guard when the task ends.

`cargo test --workspace` was green here, including `p6_imap_idle_unsolicited_exists`. I also ran a live scrape against `run_imap_listener` / `run_smtp_listener`: idle 0, one held IMAP + one held SMTP → `maddy_conns_active{module="imap"} 1` and `smtp 1`, drop → 0. The existing openmetrics e2e does not cover that path (it builds `SmtpSession` itself). A listener-level scrape in that test would be enough to lock the wiring; not a merge blocker.

PR #165, opened 2026-09-07
feat(cli): add `madmail monitor` for live connection and throughput stats

mamaru-a:
Step 4 of #19: the CLI. Top of a two-PR stack, rebased onto 2.27.0:

| | | |
|---|---|---|
| themadorg/madmail#164 | `maddy_conns_active` gauge (step 1) | base `main` |
| **this PR** | `madmail monitor` (step 4) | base `feat/19-conn-metrics` |

This is based on `feat/19-conn-metrics`, not `main`, so the diff here is only this PR's own four commits. **Merge order:** #164 first, then retarget this PR to `main` — GitHub will not do that automatically, because #164's head is a fork branch rather than the base branch this PR points at.

How to retarget depends on how #164 lands. If it is **merge-committed**, its commit is in `main`'s history and a plain `gh pr edit 165 --base main` is enough. If it is **squash-merged**, `60aed77` never reaches `main`, and retargeting alone would show the gauge changes twice — this branch has to be replayed onto the squash first:

```bash
git fetch upstream
git rebase --onto upstream/main 60aed77 feat/19-monitor-cli
git push -f origin feat/19-monitor-cli
gh pr edit 165 --repo themadorg/madmail --base main
```

```
$ madmail monitor --interval 1
Scraping http://127.0.0.1:9749/metrics every 1s
  imap   smtp submission   total     msg/s aborted/s   queue
     4      2          0       6         -         -       3
     4      2          0       6     19.88      0.00       3
     4      2          0       6     19.91      0.00       3
```

## Commits

1. `feat(cli): add madmail monitor …` — the command, plus an exact-name metrics parser.
2. `fix(cli): enable the openmetrics endpoint in the installer` — see below.
3. `test(cli): run monitor against a live openmetrics listener`
4. `docs(cli): document madmail monitor` — guide page, JSON schema, index and status rows.

## Why scrape `/metrics` instead of adding a control socket

The existing `ctl` commands are offline — they read the config and open the DB in-process — so none of them can see a running server's state. Rather than introduce a control socket for this one command, `monitor` reads the `openmetrics` endpoint the server already serves. The CLI, a Prometheus scrape and any future dashboard then read the same numbers, and there is no second code path to keep in sync. That is also the issue's "Data Export Readiness" item, satisfied by construction.

## The installer now enables the endpoint — on 9749, not 9100

The installer used to write the `openmetrics` block commented out, so on a default install `monitor` could only report that the endpoint was off. It now writes:

```
openmetrics tcp://127.0.0.1:9749 {
}
```

The old commented example used 9100. That is node_exporter's default port, and the supervisor runs `preflight_listen_addrs` on the openmetrics address at boot and propagates the failure — so enabling it on 9100 would make a fresh install fail to start on any host already running node_exporter. 9749 avoids that and matches what the error message and docs already say. It stays on loopback because the endpoint has no authentication. This only affects newly generated configs; existing ones are untouched.

If you would rather keep it opt-in, it is a one-line revert of commit 2, and the error message already names the exact line to add:

```
Error: configuration error: openmetrics endpoint is not enabled; add `openmetrics tcp://127.0.0.1:9749 { }` to the config (bind it to loopback - it has no authentication), or pass --addr
```

## Decisions worth reviewing

- **Rates are client-side**, `Δcounter / Δt` across two scrapes — the sliding window the issue asks for, without a ring buffer on the server.
- **Wall-clock `Δt`**, not the nominal interval, so scrape latency does not skew readings.
- **A counter going backwards means the server restarted**, and reports `0` for that interval, as Prometheus does for a counter reset. Otherwise a restart would show as a large negative rate.
- **The first row prints immediately with `-` for rates**, so `monitor --count 1 --json` returns current counts straight away for scripting.
- **`--json` emits one envelope per sample** (NDJSON).
- **A wildcard `openmetrics_listen`** (`0.0.0.0`, `[::]`) is a bind address, not a dial address, and is rewritten to loopback.
- **5s HTTP timeout, async reqwest client** — `reqwest::blocking` builds its own runtime and panics under the one `dispatch` runs on (see the comment at `dispatch.rs:45`).

## Metrics parser

`sample_value` in `chatmail-metrics` matched on `line.starts_with(name)`, so `maddy_smtp_started` would have matched samples of `maddy_smtp_started_transactions`. It is now an exact name match (the name must be followed by `{` or whitespace), exposed as `samples()`, which returns every labelled sample so the CLI can sum across modules. `sample_value` wraps it and keeps its behaviour and callers.

## Verification

Run on macOS against this branch rebased onto 2.27.0 (which includes upstream's own `statvfs` fix, so the workspace builds there without #166, now closed):

- `RUSTFLAGS="-D warnings" cargo check --workspace --all-targets`, `cargo clippy --workspace --all-targets -- -D warnings`, `cargo fmt --all --check`, `RUSTDOCFLAGS="-D warnings" cargo doc --no-deps` — all clean.
- `cargo test -p chatmail -p chatmail-metrics -p chatmail-config -p chatmail-smtp` — all pass, 0 failures.
- **End-to-end** (`crates/chatmail/tests/monitor_test.rs`): starts a real openmetrics listener, holds a connection guard in the test process, and runs the actual `madmail monitor --json --count 1` binary against it. It must report `imap: 1` from that live gauge — covering the gauge, the exporter, HTTP and the parser together — plus `smtp`/`submission` present as `0` and no rate on the first sample.
- **Installer round-trip**: the generated config is parsed back with `parse_maddy_config` and must yield `openmetrics_listen == 127.0.0.1:9749`, so a malformed block fails the test rather than just a missing string.
- Unit tests in `monitor.rs` cover scrape parsing, missing metrics reading as 0, rate including counter reset and zero `Δt`, wildcard address rewriting, and `--addr`/config/error precedence; one in `metrics.rs` covers exact name matching.

What is still not exercised is a full `madmail run` boot followed by `monitor`; the e2e test runs the real exporter and the real CLI binary, but not the supervisor that wires them together.

### DIFF
--- a/crates/chatmail-config/src/cli.rs
+++ b/crates/chatmail-config/src/cli.rs
@@ -192,6 +192,21 @@ pub enum Command {
     /// Migrate submission PGP policy in config.
     #[command(name = "migrate-pgp-config")]
     MigratePgpConfig,
+    /// Live server metrics (connections and message throughput).
+    ///
+    /// Polls the `openmetrics` endpoint, so that endpoint has to be enabled in
+    /// the config (`openmetrics tcp://127.0.0.1:9749 { }`).
+    Monitor {
+        /// Seconds between samples.
+        #[arg(long, short = 'n', default_value_t = 2)]
+        interval: u64,
+        /// Stop after this many samples (default: run until interrupted).
+        #[arg(long, short = 'c')]
+        count: Option<u64>,
+        /// Override the scrape address (default: `openmetrics` from the config).
+        #[arg(long, value_name = "HOST:PORT")]
+        addr: Option<String>,
+    },
     /// Show server status (connections, users, uptime).
     Status {
         /// Per-port breakdown.
--- a/crates/chatmail-metrics/src/lib.rs
+++ b/crates/chatmail-metrics/src/lib.rs
@@ -22,7 +22,7 @@ mod server;
 
 pub use metrics::{
     conn_guard, exposition_text, init_metrics, record_smtp_aborted, record_smtp_completed,
-    record_smtp_failed_command, record_smtp_failed_login, record_smtp_started, set_queue_length,
-    ConnGuard,
+    record_smtp_failed_command, record_smtp_failed_login, record_smtp_started, sample_value,
+    samples, set_queue_length, ConnGuard,
 };
 pub use server::run_openmetrics_listener;
--- a/crates/chatmail-metrics/src/metrics.rs
+++ b/crates/chatmail-metrics/src/metrics.rs
@@ -191,22 +191,48 @@ pub(crate) fn gather_bytes() -> Result<Vec<u8>, prometheus::Error> {
 }
 
 /// Parse a counter/gauge sample from Prometheus text exposition.
-#[cfg(test)]
-pub fn sample_value(body: &str, metric_name: &str, label_selector: &str) -> Option<f64> {
+/// Every sample of `metric_name` in a Prometheus text exposition, as
+/// `(label set, value)` - the label set is the raw text between the braces, or
+/// empty for an unlabelled metric.
+///
+/// The name must match exactly: asking for `maddy_smtp_started` does not return
+/// samples of `maddy_smtp_started_transactions`.
+pub fn samples(body: &str, metric_name: &str) -> Vec<(String, f64)> {
+    let mut out = Vec::new();
     for line in body.lines() {
-        if line.starts_with('#') || line.is_empty() {
+        let line = line.trim();
+        if line.is_empty() || line.starts_with('#') {
             continue;
         }
-        if !line.starts_with(metric_name) {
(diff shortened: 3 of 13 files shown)


KamyarAT:
Looked at this on its own commit, with the stack in the description.

The shape is right: scrape `/metrics`, `CtlOut`, exact metric names, wildcard bind rewritten to loopback, first row with `-` rates, backwards counter treated as a reset. Piped `--json` streams (`println!` flushes on newline). A failed scrape exiting the process is fine for a foreground CLI.

The one gap I’d still want on this PR is operator docs: `docs/guide/cli/monitor.md` (openmetrics has to be on, loopback, first-row `-` rates) and a `json-output.md` section that says this command is NDJSON (one envelope per sample; `--count 1` is the single-object form). Completions already come from clap.

Nit: Prometheus `rate()` does not report 0 on a reset; it includes the post-restart increase. Returning 0 here is fine — just don’t cite Prom for that.

PR #167, opened 2026-09-08
feat(listeners): serve IMAP and submission on the HTTPS port via ALPN

mamaru-a:
## Summary

Lets one TCP port serve both the website and mail. Port 443 now carries HTTPS, IMAP and SMTP submission on a single listener, demultiplexed by the ALPN token the client offered — so Delta Chat still connects on networks that block 993 and 465.

```
chatmail tls://0.0.0.0:443 {
    alpn_imap imap
    alpn_smtp smtp
}
```

**This is on for existing servers after upgrade.** The installer has emitted these directives since day one, but the config parser dropped them and no listener implemented ALPN, so they were dead config. This PR makes them live. Any server installed with chatmail enabled gains IMAP and submission on 443 with no config edit.

### Why the demux lives here and not in a proxy

Upstream chatmail does this in nginx's `stream` module with `ssl_preread`, passing raw TCP through to Dovecot and Postfix. Madmail is a single binary with no proxy in front, so it terminates TLS once and dispatches on the protocol rustls negotiated.

That is also the more defensive design. Because we terminate, `alpn_protocols` is set and rustls answers a client whose offers do not intersect ours with a fatal `no_application_protocol` — the ALPACA countermeasure RFC 9325 §3.8 requires. A passthrough demux cannot enforce it: the backend never negotiates ALPN at all.

### Routing rules, and why

- **ALPN only, never SNI.** SNI names a host, not a protocol, and a browser sets it from the URL bar. Routing on SNI would send `https://imap.example.org/` into the IMAP parser with no attacker involved.
- **No ALPN means HTTPS.** There is no byte-sniffing fallback. IMAP and SMTP are server-speaks-first while HTTP is client-speaks-first, so no first byte separates them without hanging one side. This matches upstream nginx's `default` branch.
- **`h2` is not advertised.** Nothing negotiates ALPN on this port today, so browsers already settle on HTTP/1.1; adding `h2` would flip the whole HTTPS surface to HTTP/2 as a side effect of enabling mail.
- **The demux takes over the existing `http_tls` slot** rather than binding 443 a second time, which would make `preflight_listen_addrs` bind the same address twice and fail startup.

### The directive value is a switch, not a token

In the Go implementation `alpn_imap` / `alpn_smtp` held a maddy **module-instance name**, which is why installed configs read `alpn_smtp submission` — that named the submission module, not an ALPN identifier.

The wire tokens are chosen by the client: `chatmail/core` sends the bare tokens `imap` and `smtp` on any non-standard port (`src/imap/client.rs`, `src/smtp/connect.rs`). So they are constants here, the directive only switches a protocol on, and a configured value that differs logs a warning naming the real token rather than being silently aliased — the drift should stay visible.

Reading the value as a wire token would have broken every existing deployment: advertising `submission` matches nothing a real client offers, and under strict ALPN that fails the handshake outright instead of quietly falling through to HTTPS.

There is no IANA-registered ALPN identifier for SMTP or submission; `imap` is registered, `smtp` is an unregistered token shared with upstream chatmail.

### Client discovery

Autoconfig now emits 443 entries for IMAP and submission, **after** the 993/465 and STARTTLS entries, so clients keep preferring the dedicated ports and only fall back to 443 where those are blocked. Each protocol is independent. `has_imap_alpn_https` had been hardcoded false behind a comment saying the feature did not exist; it now reads the running configuration, reaching the www handler through `ListenerPortsStore` — the only channel from the supervisor to that route.

`chatmail status` also parses `alpn_smtp`, so submission on 443 is no longer invisible next to the IMAP row it already printed.

### Scope — what cannot move

- **Inbound MX on 25.** MX records name a host, not a port. Unmovable.
- **STARTTLS on 143 and 587.** Their first bytes are a plaintext server greeting, not a ClientHello, so they cannot ride a TLS-first port. Only implicit TLS (993/465 semantics) can.
- **Port 80.** No ClientHello to route on, and cleartext mail is obsolete per RFC 8314. Left as ACME http-01 plus redirect.

### Not a censorship-circumvention claim

ALPN tokens travel in the plaintext ClientHello, so `imap` on the wire is as visible as port 993 — and chatmail/core#8674 argues the unconditional token is itself a DPI fingerprint, since browsers only ever offer h3/h2/http1.1. This solves port blocking (hotel and corporate wifi), not deep packet inspection. It also requires L4 passthrough, so it is not compatible with a TLS-terminating CDN in front of 443.

## Type of change

## Area

- [x] SMTP
- [x] IMAP
- [x] Configuration / CLI

## Testing

16 new tests, written before the code they cover. The two integration tests drive real TLS handshakes against a real listener on an ephemeral port rather than asserting on mocks:

- `p12_it01_one_port_serves_imap_smtp_and_http` — connects three times to one listener and asserts the IMAP greeting for ALPN `imap`, the SMTP banner for ALPN `smtp`, and an HTTP response for a client sending no ALPN.
- `p12_it02_unoffered_alpn_is_refused` — a client offering only `h2` is rejected at the handshake, locking the strict-ALPN property.
- `p12_it03_mail_autoconfig_advertises_https_alpn_when_configured` — end-to-end through the served XML.
- `p12_ut16_legacy_directive_value_never_reaches_the_wire` — a config reading `alpn_smtp submission` still puts `smtp` on the wire.

Plus unit coverage of the routing table, advertised-token order, per-protocol independence, and the parser.

**Manual verification:**

```text
cargo test --workspace     # green except p6_imap_idle_unsolicited_exists (see below)
```

Two notes for reviewers:

- `p6_imap_idle_unsolicited_exists` in `chatmail-imap` is **pre-existing flaky** — it passes roughly one run in three on an unmodified tree and is untouched by this PR.
- On macOS the workspace does not build without #166 (`statvfs` field widths). That is pre-existing on `main` and deliberately left to that PR rather than duplicated here; local verification was run with #166's change applied to the working tree.

## Documentation

`13-configuration.md` documented this feature as unimplemented; it now describes the config surface, the ALPN-only rule, strict-ALPN rationale, the legacy directive value, and the scope limits. Renamed autoconfig tests are corrected in `16-testing.md` and `10-webimap.md`.

## Security & privacy

Two properties are locked by tests: a connection with no ALPN is never routed to a mail parser, and a client whose ALPN offers do not intersect ours is refused at the handshake. Both are ALPACA mitigations.

**Known residual, not introduced here:** 993 and 465 still advertise no ALPN while sharing the website's certificate, so they remain substitute servers for the web origin regardless of how strict 443 is — TLS cannot bind a connection to an intended port. Pre-existing; worth a follow-up that gives those listeners their own ALPN lists.

## Commits

- [x] `feat:` new feature — single squashed commit

## Checklist

Parity note: v1 implemented this demux in its Go maddy fork, dispatching on module instances. This reimplements the behaviour in Rust with the same wire tokens, and additionally enforces strict ALPN, which v1's passthrough could not.


KamyarAT (approved):
The demux itself looks right: ALPN only (not SNI), no ALPN → HTTPS, `alpn_smtp submission` is a switch and the wire token stays `smtp`, demux takes the existing `http_tls` slot so preflight does not bind 443 twice, 443 mail is submission (AUTH) not MX. The live tests passed here (`p12_it01` three protocols on one port, `p12_it02` `h2`-only refused, `p12_it03` autoconfig). `cargo test --workspace` was green.

Known leftovers, not merge blockers once rebased: this is on for existing configs that already have the directives; a TLS-terminating CDN in front of 443 cannot see the mail tokens; autoconfig 443 is only useful to Delta Chat until #170 (stock clients send no ALPN and would hang); 993/465 still have no ALPN (#169).

KamyarAT:
@mamaru-a 
GitHub marks this conflicting with `main`. Please do not merge `main` into the branch; that is what blows up `dispatch.rs`. #164 and #165 already landed as different commits, and this branch still contains the old copies.

Replay **only** the ALPN commit onto current `main`:

```bash
git fetch upstream   # or origin, whichever is themadorg/madmail
git rebase --onto upstream/main 88b9046e feat/shared-ports-web-and-email
git push -f
```

`88b9046e` is the parent of `a10a7bd`. After that, `git log upstream/main..HEAD` should be a single commit, and the diff should not include `monitor.rs` or the gauge wiring.

I trial-rebased that locally; it applied clean. #169 and #170 should wait until this rebased branch is on `main`, then do the same with their own tip commit.

mamaru-a:

PR #168, opened 2026-09-08
fix(federation): drop the unreachable SMTP-over-443 delivery retry

mamaru-a:
## Summary

Removes the implicit-TLS retry on a peer's port 443 from outbound SMTP federation. It could never deliver a message, and it made every federation failure take 30 seconds longer than it needed to.

`deliver()` tried `host:25` and, on any failure, retried `host:443` with implicit TLS. Two independent reasons that retry is dead code:

1. **It reaches the wrong service and then blocks.** The federation SMTP client never sets `alpn_protocols`, so against a relay that demultiplexes 443 by ALPN the connection lands in the `default` branch — the internal HTTPS server. `deliver_implicit_tls` writes nothing and immediately waits for a server-first `220` banner. An HTTP server never speaks first, so the connection blocks until the 30s timeout and only then fails. It did not fail fast.

2. **Even with the right ALPN token it would be rejected.** Relays route the `smtp` token to their submission service, which is configured `smtpd_sasl_auth_enable=yes` with `smtpd_relay_restrictions=permit_sasl_authenticated,reject`. Port 443 is a *client submission* path there, never an MX path.

Relays do listen on 25 for inbound MX mail, and the documented relay-to-relay alternative is an HTTPS `POST /mxdeliv` — which this repo already implements and already tries first (`transport.rs` runs HTTPS, then HTTP, then SMTP). So the retry never contributed a delivery, only latency on the failure path.

Port 25 is now the only SMTP port tried, and its error is returned directly rather than wrapped in a second failure that hides it.

Also corrects two doc comments that asserted the opposite and would invite someone to add this back:

- `federation_smtp.rs` — "required for classic Chatmail relays (`nine.testrun.org`, etc.) that expose SMTP only on :443". They run Postfix on 25 for MX.
- `transport.rs` — "Peers like nine.testrun.org often only accept real SMTP on :443". The surrounding comment's actual point (don't treat a 4xx from a missing `/mxdeliv` as permanent) is kept.

## Type of change

## Area

- [x] SMTP

## Testing

Test written first, and it failed for the right reason before the change:

```text
must not retry on :443, got: smtp :25 failed; smtp :443 tls: smtp tls connect:
failed to lookup address information: nodename nor servname provided, or not known
```

`p7_ut07_no_implicit_tls_retry_on_443` drives `deliver()` at a `.invalid` host — reserved by RFC 2606, so it never resolves and the test needs no network — and asserts the resulting error never mentions 443. The assertion is on error content rather than elapsed time, so it is not timing-flaky.

**Manual verification:**

```text
cargo test -p chatmail-delivery
cargo test --workspace
```

Two notes for reviewers:

- `p6_imap_idle_unsolicited_exists` in `chatmail-imap` is **pre-existing flaky** — roughly one pass in three on an unmodified tree, unrelated to this change.
- On macOS the workspace does not build without #166 (`statvfs` field widths); that is pre-existing on `main`. Local verification here was run with #166's change applied to the working tree and then reverted, so it is not part of this diff.

## Documentation

`07-federation.md` already listed port 25 as the last-resort SMTP step and never documented the 443 retry — the code was the outlier. Added an explicit note on why 443 is not tried, so it does not get reintroduced.

## Security & privacy

- [x] Not security-sensitive

Removes an outbound code path; adds none. Worth noting it deletes the only caller of the federation TLS client that ran *without* STARTTLS negotiation, but the `SkipServerVerification` verifier is unchanged and still used by the STARTTLS path.

## Commits

- [x] `fix:` bug fix

## Checklist

Parity note: no delivery behaviour is lost. Any peer reachable before is still reachable via `/mxdeliv` or port 25; only a path that could not complete a transaction is gone.

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -877,6 +877,7 @@ dependencies = [
  "chatmail-config",
  "chatmail-db",
  "chatmail-iroh",
+ "chatmail-metrics",
  "chatmail-pgp",
  "chatmail-push",
  "chatmail-state",
--- a/crates/chatmail-config/src/cli.rs
+++ b/crates/chatmail-config/src/cli.rs
@@ -192,6 +192,21 @@ pub enum Command {
     /// Migrate submission PGP policy in config.
     #[command(name = "migrate-pgp-config")]
     MigratePgpConfig,
+    /// Live server metrics (connections and message throughput).
+    ///
+    /// Polls the `openmetrics` endpoint, so that endpoint has to be enabled in
+    /// the config (`openmetrics tcp://127.0.0.1:9749 { }`).
+    Monitor {
+        /// Seconds between samples.
+        #[arg(long, short = 'n', default_value_t = 2)]
+        interval: u64,
+        /// Stop after this many samples (default: run until interrupted).
+        #[arg(long, short = 'c')]
+        count: Option<u64>,
+        /// Override the scrape address (default: `openmetrics` from the config).
+        #[arg(long, value_name = "HOST:PORT")]
+        addr: Option<String>,
+    },
     /// Show server status (connections, users, uptime).
     Status {
         /// Per-port breakdown.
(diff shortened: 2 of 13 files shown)


mamaru-a:

PR #169, opened 2026-09-08
fix(security): advertise strict ALPN on the dedicated mail TLS ports

mamaru-a:
## Summary

> **Depends on #167 — merge that first.** This branch is cut from `feat/shared-ports-web-and-email` because it reuses `load_server_config_with_alpn` and the `ALPN_IMAP` / `ALPN_SMTP` constants introduced there. GitHub cannot base a PR on a branch that exists only in the fork, so this targets `main` and its diff currently includes #167's commit as well. Once #167 merges, the diff here collapses to this branch's own commit — the last one, `fix(security): advertise strict ALPN on the dedicated mail TLS ports`. Review that one.

Makes the dedicated implicit-TLS mail ports advertise strict ALPN, closing the cross-port half of the ALPACA exposure that #167 only closed on 443.

Every TLS listener here presents the same certificate, because there is one hostname. That removes ALPACA's certificate-separation countermeasure by construction. And TLS does not bind a connection to an intended port — so 993 accepting any client remained a valid substitute server for the web origin however strictly 443 enforces ALPN. Strict ALPN is the only TLS-layer defence left, and RFC 9325 (BCP 195) §3.8 recommends it.

| Listener | Advertised ALPN | Change |
|---|---|---|
| IMAP implicit TLS (993) | `imap` | **new** |
| Submission implicit TLS (465) | `smtp` | **new** |
| HTTPS sharing mail (443) | enabled mail tokens + `http/1.1` | from #167 |
| HTTPS alone (443) | none | unchanged |
| STARTTLS (143 / 587), inbound SMTP (25) | none | unchanged |

`load_tls_config` now returns a `MailTlsConfigs` of three configs built from one certificate instead of a single shared `Arc`, so each listener gets the right one.

### Why this does not lock anyone out

rustls sends `no_application_protocol` **only** when the client offered ALPN and nothing overlapped. A client that sends no ALPN extension at all negotiates none and connects normally. That is what keeps Thunderbird and Apple Mail working — neither sends ALPN — and Delta Chat omits it on the standard ports too, sending a token only on non-standard ones.

That distinction is the entire safety argument for this change and it is not visible from the config, so it is verified against a real listener rather than assumed. `p12_it04_alpn_less_client_still_reaches_imap` drives an ALPN-less client through a real handshake to the `* OK` greeting on an ALPN-restricted listener, and also checks a client offering `imap` negotiates it.

### Deliberate limits

- **STARTTLS ports advertise nothing.** 143 and 587 negotiate TLS after a plaintext greeting, so the client has no ClientHello to carry ALPN; restricting that config would refuse every STARTTLS client. Explicitly not touched, and asserted by test.
- **Port 25 unchanged**, for the same reason plus MTA compatibility.
- **An ALPN-less client stays unattributable.** On the shared 443 port such a connection is served as HTTPS rather than guessed into a mail parser — unchanged from #167.

## Type of change

## Area

- [x] SMTP
- [x] IMAP
- [x] Configuration / CLI

## Testing

Four tests, the wiring one written first and failing on the missing builder before the change:

- `p12_ut17_dedicated_mail_ports_get_their_own_alpn` — locks the per-listener token lists, including that the STARTTLS config stays empty. This is the one that catches a future regression in the supervisor wiring.
- `p12_it04_alpn_less_client_still_reaches_imap` — real listener, real handshake: no ALPN still reaches the IMAP greeting; `imap` negotiates.
- `p12_it05_browser_alpn_refused_on_imap_port` — a browser's `h2, http/1.1` offer is refused on 993.

I also reviewed the wiring diff line by line to confirm port 25 and the plain submission/IMAP listeners still receive `tls = None` with the ALPN-free config as their STARTTLS upgrade, rather than accidentally being handed an implicit-TLS config.

**Manual verification:**

```text
cargo test -p chatmail --lib p12_
cargo test --workspace
```

Two notes for reviewers:

- `p6_imap_idle_unsolicited_exists` in `chatmail-imap` is **pre-existing flaky** — about one pass in three on an unmodified tree, unrelated to this change.
- On macOS the workspace does not build without #166 (`statvfs` field widths), which is pre-existing on `main`. Local verification was run with #166's change applied to the working tree and then reverted, so it is not in this diff.

## Documentation

New "Cross-Protocol Attacks (ALPACA)" section in `12-security.md` with the per-listener table, why certificate separation is unavailable here, why the STARTTLS ports are exempt, and the tests that lock it. RFC 7301 and RFC 9325 added to the RFC index.

## Security & privacy

- [x] Reviewed error messages for information leakage

The rejection is a standard TLS alert from rustls; no application-level message is emitted, so nothing new is disclosed.

Residual worth stating: this raises the cost of a cross-protocol attack but does not eliminate every path. ALPACA also covers confusion between two services running the *same* protocol on different ports under the same name, which ALPN cannot distinguish.

## Commits

- [x] `fix:` bug fix

## Checklist

Parity note: v1 advertised no ALPN on any listener. This is an intentional divergence — a hardening measure v1's architecture could not express, since its 443 demux was an nginx passthrough that never negotiated ALPN.


mamaru-a:

PR #170, opened 2026-09-08
feat(listeners): identify shared-port mail by hostname and first bytes

mamaru-a:
## Summary

> **Depends on #167 and #169 — merge those first.** Cut from `security/strict-alpn-mail-ports`, which is itself cut from `feat/shared-ports-web-and-email`. GitHub cannot base a PR on a branch that exists only in the fork, so this targets `main` and its diff currently includes both parents' commits. Once they merge, the diff here collapses to this branch's own commit — the last one, `feat(listeners): identify shared-port mail by hostname and first bytes`.

Adds two more ways to identify a connection on the shared HTTPS port, so it stops being Delta-Chat-only.

Today ALPN is the sole signal, and Delta Chat is the only client that sends the `imap` / `smtp` tokens. **Thunderbird, Apple Mail and K-9 send no ALPN at all**, so under #167 they always land on HTTPS and the shared port is useless to them. Two more signals now sit *behind* ALPN:

| # | Signal | When | Identifies |
|---|--------|------|-----------|
| 1 | **ALPN** (ClientHello) | always | Delta Chat, browsers |
| 2 | **SNI** (same ClientHello) | only if no ALPN was offered | any stock mail client pointed at `imap.example.org:443` |
| 3 | **First bytes** (post-handshake) | only if neither | early `EHLO`/`HELO`, pipelined IMAP command, HTTP request line |

Anything still unidentified — silence included — is served as HTTPS rather than guessed into a mail parser.

### The order is the security property, not a preference

I argued against SNI routing in #167 and that argument still holds — *for SNI instead of ALPN*. SNI names a host, not a protocol, and a browser sets it from its URL bar, so consulting SNI **first** would send a visitor to `https://imap.example.org/` straight into the IMAP parser with no attacker involved.

Checking ALPN first closes that. Every browser offers ALPN in order to negotiate HTTP/2, so a browser is always decided at step 1 and can never reach steps 2 or 3. `p12_it07_alpn_outranks_sni` locks this: a client presenting `imap.localhost` as SNI while offering `http/1.1` gets HTTP.

### On EHLO

Worth being explicit, since it was specifically requested: **EHLO cannot route a connection by itself.** By the time a client is allowed to send it, the server has already had to decide the protocol and send `220`. It is only meaningful as an *early-talker* signal inside the peek — a client that speaks before being greeted. Conforming clients don't, so that branch fires rarely; it is the peek as a whole (EHLO / IMAP tag / HTTP verb / silence) that carries the value. Documented as such rather than oversold.

### Hostname configuration

```
chatmail tls://0.0.0.0:443 {
    alpn_imap imap
    alpn_smtp smtp
    sni_imap  mail.example.org      # optional
    sni_smtp  send.example.org      # optional
}
```

Unset, the conventional `imap.` / `smtp.` first labels select mail. Setting one **replaces** the prefix for that protocol rather than adding to it — an operator picking a neutral name does so precisely so `imap.` is not left as a routable marker on their server.

**The certificate must cover whichever hostname clients use**, since the client validates it before any routing happens. That is the real constraint on this feature and it is called out in the docs.

### Implementation notes

- Reading SNI before choosing a config needs the ClientHello, so the listener moves from `TlsAcceptor` to `LazyConfigAcceptor`. The `ClientHello` borrows the handshake, so the decision is taken in an inner scope before `into_stream` consumes it.
- The session entry points become generic over the stream, so the peek path can hand on a reader that already buffered the client's first bytes. Buffering is applied **only** on that path — the ALPN and SNI paths pass the raw TLS stream, so nothing is added to the hot path (this matters: federation `/mxdeliv` bodies reach 70 MB).
- `fill_buf` is cancel-safe, so timing out the peek cannot consume bytes.
- Strict ALPN from #167 is unchanged: a client offering ALPN that does not intersect ours still gets `no_application_protocol`.

## Type of change

## Area

- [x] SMTP
- [x] IMAP
- [x] Configuration / CLI

## Testing

Ten new tests, written before the code. 23 P12 tests pass in total.

Integration, over real TLS against a real listener:

- `p12_it06_sni_identifies_mail_without_alpn` — an ALPN-less client presenting `imap.localhost` reaches `* OK`; `smtp.localhost` reaches `220`.
- `p12_it07_alpn_outranks_sni` — the security ordering above.
- `p12_it08_first_bytes_identify_when_nothing_else_does` — an early `EHLO` on a neutral hostname reaches submission; an HTTP request line reaches HTTPS.

Unit, on the decision logic: prefix routing, case/trailing-dot normalisation, configured name retiring the prefix, a disabled protocol being unreachable by hostname, and the first-byte classifier including its refusals (`GETX`, an unknown IMAP command, and a TLS record header all return no verdict).

**Manual verification:**

```text
cargo test -p chatmail --lib p12_
cargo test --workspace
```

Two notes for reviewers:

- `p6_imap_idle_unsolicited_exists` in `chatmail-imap` is **pre-existing flaky** — about one pass in three on an unmodified tree, unrelated to this change.
- On macOS the workspace does not build without #166 (`statvfs` field widths), pre-existing on `main`. Verification was run with #166 applied to the working tree and reverted before committing, so it is not in this diff.

## Documentation

`13-configuration.md` gets the three-signal order, the `sni_imap` / `sni_smtp` surface, the certificate constraint, and why `EHLO` cannot route on its own. `12-security.md` records why ALPN-before-SNI is the safe ordering. The `shared_listener` module doc is rewritten — it previously asserted "ALPN only, never SNI", which this change makes false.

## Security & privacy

- [x] Reviewed error messages for information leakage

The ALPN-before-SNI ordering is the property, and it is locked by test rather than by comment.

Residual, stated plainly: a **non-browser** TLS client that sends no ALPN and connects to the mail hostname reaches a mail parser. That is the intended user of the feature — a stock mail client — and it is a narrower surface than SNI-first routing, which a plain browser URL could reach. Operators who do not want it can leave `sni_imap` / `sni_smtp` unset and keep a non-`imap.`-prefixed hostname, which restores ALPN-only behaviour.

## Commits

- [x] `feat:` new feature

## Checklist

Parity note: v1's nginx `ssl_preread` demux could read SNI too, but its config only mapped ALPN. This is a superset, with the ordering made explicit.


mamaru-a:

